Issue #33, opened 2022-11-04
Ngen-cal strategy field unexpectedly expressed as enum in json schema


aaraney:
https://github.com/NOAA-OWP/ngen-cal/blob/fdca5620efc1f021af9ce044aab2e4852a81792a/python/ngen_cal/ngen.py#L240

From the src, it looks like what this should expand to is the following:

```python
class NgenExplicit(NgenBase):
    
    strategy: Literal["explicit"]
```

However, when you export the model as a json schema using pydantic, you get the following:

```json5
      // ...
      "strategy": {
          "title": "Strategy",
          "enum": [
            "explicit"
          ],
          "type": "string"
        }
      // ...
```

I think the following should remedy this:

```python
class NgenExplicit(NgenBase):
    
    strategy: Literal[NgenStrategy.explicit.value] = NgenStrategy.explicit.value

```

aaraney:
This applies to `NgenExplicit`, `NgenIndependent`, and `NgenUniform`.

aaraney:
Now that I look back on this, im not sure _why_ this is a problem. I think the json schema has to be this way even if you provide `Literal["explicit"]` as the field type hint. `Literal` is in effect an enum type.

Issue #32, opened 2022-11-04
Typo in formulation model constraint


aaraney:
https://github.com/NOAA-OWP/ngen-cal/blob/fdca5620efc1f021af9ce044aab2e4852a81792a/python/ngen_conf/src/ngen/config/formulation.py#L8

This should be `discriminator`. Likewise, the discriminator should be `model_type_name` not `model_name`. `model_type_name` is the alias for `model_name`, meaning `model_type_name` is the key in the configuration file.

Edit:

While this was an issue, by correctly spelling `discriminator`, this forced all `model_type_name` fields to be type `Literal`. This is okay in most cases, however it is not desirable for `MultiBMI`. `MultiBMI`'s `model_type_name` field should not be constrained like the other BMI model pydantic model types.

However, there still needs to be a way to determine the correct pydantic model variant from the union of formulations, `KnownFormulations`. The `name` and `model_type_name` fields are used for this purpose. `name` corresponds the bmi adapter name used by that model (e.g. `bmi_fortran`) and `model_type_name` roughly corresponds to the model name (e.g. `NoahOWP`). For non-multi bmi models, the `name` _and_ `model_type_name` are constant values (e.g. `bmi_fortran` and `NoahOWP` for the `NoahOWP` pydantic model). In the case of the `MultiBMI` pydantic model, only _name_ is a constant value, `bmi_multi`, `model_type_name` can be any valid string.

However, it feels tedious that it would be required to specify these constants when programmatically creating a for example, a `NoahOWP` instance. For this reason, these constants _are not required_ when creating a `KnownFormulations` type. Likewise, it is possible to pass a `KnownFormulations` instance when creating a `Formulation` object. However, to deserialized into `Formulation` object, it _is_ required that `name` _and_ `model_type_name` are present for non-multi bmi types and ,potentially only, `name` it equals bmi_multi`.

aaraney:
I can get a PR in for the on Monday. If I forget, just ping me.

hellkite500:
@aaraney ping?

PR #34, opened 2022-11-14
Fix minor bugs with default values and specificity in ngen.cal and ngen.config

### LINKED ISSUE #31: ngen.cal Objective Enum needs type constraints
https://github.com/NOAA-OWP/ngen-cal/blob/fdca5620efc1f021af9ce044aab2e4852a81792a/python/ngen_cal/strategy.py#L19

This enum should be a `class Objective(str, Enum)`. 

When `Objective` is used as a type in the [`Estimation`](https://github.com/NOAA-OWP/ngen-cal/blob/fdca5620efc1f021af9ce044aab2e4852a81792a/python/ngen_cal/strategy.py#L61) pydantic model, the resulting `json_schema` does not qualify that that the `objective` field is constrained to string types.

```json5
  "definitions": {
    // ...
    "Objective": {
      "title": "Objective",
      "description": "Enumeration of supported search algorithms\n\n    ",
      // missing "type": "string",
      "enum": [
        "custom",
        "kling_gupta",
        "nnse",
        "single_peak",
        "volume"
      ]
    },
    "Estimation": {
      "title": "Estimation",
      "description": "Estimation strategy for defining parameter estimation",
      "type": "object",
      "properties": {
        // ...
        "objective": {
          "title": "Objective",
          "default": "custom",
          "anyOf": [
            {
              "$ref": "#/definitions/Objective"
            },
            // ...
          ]
        },
    // ...
```

aaraney:
Fixes minor issues in cal and config's pydantic model fields. This does not introduce any breaking changes.

fixes #31, #32, #33

### Changes

- `ngen.cal` `Objective` enum now properly subclasses `str`. This fixes pydantic models' json schemas that use `Objective` as a field type. See #31 for impacts.
- `ngen.cal` calibration strategy types: `NgenExplicit`, `NgenIndependent`, and `NgenUniform`, now have default `strategy` field values.

### Changes

- Fixed field discriminator issue in `ngen.config`'s `Formulation` model `params` field. 
Note, during `Formulation` object creation if the `params` field is deserialized (e.g. `params`'s value is a dictionary), the `name` field is required. If `name` *is not* 'bmi_multi', the `model_type_name` field is also required. Neither are required if a concrete known formulation instance is provided (see #32).

## Testing

1. Added tests to verify `NgenExplicit`, `NgenIndependent`, and `NgenUniform`  `strategy` field default value.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

- [x] Linux
- [x] MacOS


aaraney:
So, the issue is in the `Formulation` pydantic model on the `params` field. Now that the discriminator field is `model_name` (alias is `model_type_name`), pydantic wants that field to be a `Literal` instead of a string. This should be a constant to for `Topmod`, `CFE`, `NoahOWP`, and `LSTM` but not `MultiBMI`, correct @hellkite500? Or can we also make it a constant, i.e. a `Literal` for `MultiBMI` as well?

hellkite500:
The multi BMI should be constant at the top level, and the sub modules would be constant in the recursive modules under there

aaraney:
> The multi BMI should be constant at the top level, and the sub modules would be constant in the recursive modules under there

So it is safe to tie `model_name` to a Literal? For concreteness, `LSTM.model_name` should always equal `LSTM`?

hellkite500:
For now I think that will work. That was the intent of the original implementation at least...

aaraney:
> For now I think that will work. That was the intent of the original implementation at least...

Just so we are 1000% on the same page, this means that all `MultiBMI` formulations will have a `model_name` of `BMIMulti`.

hellkite500:
>Just so we are 1000% on the same page, this means that all MultiBMI formulations will have a model_name of BMIMulti
As long as that only applies to the container, not the sub formulations, then I don't see a problem with that. (though we can decide if it should be BMIMulti or MultiBMI...haha)

aaraney:
@robertbartel was having some related issues [over here](https://github.com/NOAA-OWP/DMOD/pull/316#issuecomment-1490912100) which made me revisit this.

> The multi BMI should be constant at the top level, and the sub modules would be constant in the recursive modules under there

In re-reading this, @hellkite500, I think I was being a little pedantic in my use of the word 'constant'. To re-phrase, is it okay to enforce that _all_ `MultiBMI` models have the same `model_name` field value? Right now, this is not the case `model_name` is being used to often convey the formulations in the modules field (i.e. `"model_type_name": "bmi_multi_noahowp_cfe"`).

robertbartel:
@aaraney, @hellkite500, I'm only just starting to dig into this, so I may not quite understand everything correctly yet.  But _ngen-config_ functionality seems to depend on using specific strings for the `model_name` of a _Formulation_ object (i.e., the `model_type_name` element(s) in realization configs).  

E.g., a BMI multi-formulation must be configured with something like this in the realization config file:

```json
   "model_type_name": "BMIMulti",
```

Based on the first few example realization configs I have looked at after beginning to review this ([see here](https://github.com/NOAA-OWP/DMOD/pull/316#issuecomment-1490912100)), that appears to be a condition that not necessarily followed in practice or in principle, at least in the provided examples.  I.e., `model_type_name` is used to provide a summary description, like `bmi_multi_noahowp_cfe`.

Additionally, the _ngen_ documentation on [BMI formulations in realization configs](https://github.com/NOAA-OWP/ngen/blob/master/doc/BMI_MODELS.md#required-parameters) only indicates that _some_ `model_type_name` value is required.  It doesn't express a need for any specific values.  It actually does something closer to the opposite, conveying that the particular value isn't even always important, despite the field itself always being required.  I haven't yet checked to see if the _ngen_ code is consistent with the documentation in that manner; perhaps @mattw-nws or @hellkite500 might verify.

This probably warrants further design and dependencies discussion, although a verification of my assessments is probably needed first.

robertbartel:
@aaraney, I rebased this branch, then fixed a few other non-_Literal_ `model_type_name` fields, and I was able to get tests passing locally for the related [dmod#316](https://github.com/NOAA-OWP/DMOD/pull/316).  I have that in my fork, under https://github.com/robertbartel/ngen-cal/tree/pr/34/rebased, if that's of any use to this issue.

aaraney:
@robertbartel, I see. This PR predates the `SLOTH` module, so it was not added in this PR. That makes sense why it was not working for you.

aaraney:
I added support for `SLOTH` and `PET` plus fixed a test that was failing.

aaraney:
The tests should fail because of #62.

aaraney:
The only [tests are that failing](https://github.com/NOAA-OWP/ngen-cal/actions/runs/5648093629/job/15299557714#step:7:45) are because of #62. Ill get in a quick PR to fix that, we can rebase this once that fix is in and re-review.

python/ngen_conf/src/ngen/config/multi.py
```diff
@@ -22,7 +23,7 @@ class MultiBMI(BMIParams, smart_union=True):
     # NOTE this is derived from the list of modules
     main_output_variable: Optional[str]
     #NOTE aliases don't propagate to subclasses, so we have to repeat the alias
-    model_name: Optional[str] = Field(alias="model_type_name")
+    model_name: str = Field(alias="model_type_name")
```
hellkite500, line 26: Could this still be optional?

python/ngen_conf/tests/data/test_config.json
```diff
@@ -6,7 +6,7 @@
                 "name": "bmi_multi", 
                 "params": {
                     "name": "bmi_multi", 
-                    "model_type_name": "NoahOWP_CFE", 
+                    "model_type_name": "BMIMulti",
```
hellkite500, line 9: Should we revert this if "BMIMulti" isn't the required literal now?

aaraney: That's fine with me. Good catch

python/ngen_conf/src/ngen/config/multi.py
aaraney: I don't think so. In the super class it is a `str`. I think the most sensible thing is to either set a default empty string or leave it as is. Thoughts?

hellkite500: well the validator `build_model_name` was using the optional `None` in order to build the name from the named nested modules.  We can set a default empty string `""` but would need to ensure that validator recognizes that and substitutes the built name.

hellkite500 (dismissed):

python/ngen_conf/src/ngen/config/multi.py
aaraney: Yeah, this is one of my gripes with `pydantic`. There is not a semantic way to signify to the caller that something is computed if it is not provided. The way that we have been doing this in DMOD is to set the default of a non-optional field to `None`. I guess in a perfect world, we would create a constant, say `COMPUTED` and use that as the default value.

Here, I just went with setting the default value to `None` and updated the root validator to always set the model name to something if it is not provided.

hellkite500 (approved):

PR #44, opened 2023-03-06
Ngen config validation

JordanLaserGit:
Adding a configuration validation python file, catchmentnexus module, and test .jsons. This allows for the ngen configuration files to be validated prior to loading costly data files.

## Additions
a configuration validation python file, catchmentnexus module, and test .jsons

-

## Removals
None
-

## Changes
None
-

## Testing

1. None. 

## Notes

-

## Todos

1) Add testing
2) Add validation logic
-

## Checklist

- [ x] PR has an informative and human-readable title
- [ x] Changes are limited to a single goal (no scope creep)
- [ x] Code can be automatically merged (no conflicts)
- [ x] Code follows project standards (link if applicable)
- [ x] Passes all existing automated tests
- [ x] Any _change_ in functionality is tested
- [ x] New functions are documented (with a description, list of inputs, and expected output)
- [ x] Placeholder code is flagged / future todos are captured in comments
- [ x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

- [x ] Linux
- [ ] MacOS


aaraney:
Hey @JordanLaserGit! Thanks for making these contributions! 

Having looked through your additions, I am curious what your thoughts are on using the [`geojson-pydantic`](https://github.com/developmentseed/geojson-pydantic/tree/main) library rather than introducing `pydantic` subtypes here that work similarly? `geojson-pydantic` fully captures the `geojson` spec using `pydantic` models which might be nice if constraints change in the future. Having looked at their docs, it appears we may only need the `CRSProperty` and `FeatProperties` code you introduced plus a `pydantic` model that captures a nexus geojson's properties.

A few other minor things, a few tests would be nice! Likewise, if you could move the test data that you added here to `tests/data` that would be great! We don't have documentation for setting up a testing environment (eeek, I need to add that), but here is a quick run down. We use [`pytest`](https://docs.pytest.org/en/7.2.x/). I would advice installing the `ngen.config` package in editable mode using `pip install -e python/ngen_conf"[develop]"`. By adding the `develop` extra requirement, `pytest` will also be installed. Once you have that installed, `cd` into `python/ngen_conf` and run `pytest` or `python -m pytest`. This will run all of the unit tests. You can create a new python module prefixed with `test_` in the `tests/` directory and put your tests there. Tests are just functions that are prefixed with `test_`. A quick look at the existing tests should get you up to speed quickly!

Thanks again for the contribution!

JordanLaserGit:
> Hey @JordanLaserGit! Thanks for making these contributions!
> 
> Having looked through your additions, I am curious what your thoughts are on using the [`geojson-pydantic`](https://github.com/developmentseed/geojson-pydantic/tree/main) library rather than introducing `pydantic` subtypes here that work similarly? `geojson-pydantic` fully captures the `geojson` spec using `pydantic` models which might be nice if constraints change in the future. Having looked at their docs, it appears we may only need the `CRSProperty` and `FeatProperties` code you introduced plus a `pydantic` model that captures a nexus geojson's properties.
> 
> A few other minor things, a few tests would be nice! Likewise, if you could move the test data that you added here to `tests/data` that would be great! We don't have documentation for setting up a testing environment (eeek, I need to add that), but here is a quick run down. We use [`pytest`](https://docs.pytest.org/en/7.2.x/). I would advice installing the `ngen.config` package in editable mode using `pip install -e python/ngen_conf"[develop]"`. By adding the `develop` extra requirement, `pytest` will also be installed. Once you have that installed, `cd` into `python/ngen_conf` and run `pytest` or `python -m pytest`. This will run all of the unit tests. You can create a new python module prefixed with `test_` in the `tests/` directory and put your tests there. Tests are just functions that are prefixed with `test_`. A quick look at the existing tests should get you up to speed quickly!
> 
> Thanks again for the contribution!

I wasn't aware of geojson-pydantic. I guess I sort of re-invented the wheel there... I have recreated the validation using geojson-pydantic. Thanks for the suggestion.

I have moved the catchment/nexus/realization test files into tests/data. I think test_config.json may need to be moved as well. Also, I moved conf_validation.py into ngen_conf/src/ngen/conf.

I have created a pytest (test_conf.py) that checks the test files. 

Thanks for your review and let me know if anything else needs to be addressed.

python/ngen_conf/tests/data/test_catchment_config.geojson
aaraney, line 1: I should have been more explicit in my first comment about moving this, so that is on me. Can you create a new directory called `hydrofabric` in the `tests/data` directory and put the hydrofabric testing data in there? It is a little nit-picky, just trying to keep all the test data organized.

python/ngen_conf/tests/data/test_nexus_config.geojson
aaraney, line 1: ditto for moving this.

aaraney (changes requested):
Thanks for making the requested changes @JordanLaserGit! As I re-reviewed this, I think that `geojson_pydantic` really simplified how this PR will accomplish it's goal. One thing that stuck out that would be nice to include here are pydantic models that capture the geojson geometry types and properties for the different hydrofabric geojson files. For example, a `nexus` file is a `FeatureCollection` of `Point` Features each with property `toid` that has a str value. In pydantic / geojson_pydantic that might look something like:

```python
from pydantic import BaseModel
from geojson_pydantic import FeatureCollection, Feature, Point


class NexusFeatureProperty(BaseModel):
  toid: str

  # NOTE: you could add a validator here to check the `toid`'s prefix.

NexusFeatureCollection = FeatureCollection[Point, NexusFeatureProperty]
```

I think it makes sense that these live in their own module in `ngen.config` something like `hydrofabric.py`(better thoughts are always wanted!)?

One other thing to note, if we are going to depend on `geojson_pydantic`, that will need to be added to `ngen.config`'s `setup.cfg` so that `pip` will install it when `ngen.config` is installed.

One last aside, `pydantic` is great, but it also has a fair number of bells and whistles as im sure you are finding out. [Here](https://docs.pydantic.dev/usage/models/#model-properties) is a listing of methods and properties that each model gets. The ones I noticed you might be interested in are `.parse_file()` and `.parse_obj()`.

python/ngen_conf/src/ngen/config/conf_validation.py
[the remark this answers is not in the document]
JordanLaserGit: I think I'm cramming a few too many things into this file. The motivation for me working on this was Zach mentioned we needed to validate ngen CLI arguments (3 config files and 2 str args for catchment/nexus sub selection). Which requires a few steps:

1) Validate that the config files themselves are valid (this is what geojson_pydantic helps with)
2) Validate that the config files work appropriately together (ensuring the toid's match between the catchment and nexus configs)
3) Validate that the catchment/nexus sub selections are valid (ensure the sub selections are actually within the configs)

conf_validation is these 3 steps packed together. I originally attempted step 1 in python/ngen_conf/src/ngen/config/catchmentnexus.py with the NGenCatchmentNexus class. I can recreate this with the geojson_pydantic library and add classes like NexusFeatureProperty.

Perhaps it makes sense then to keep steps 2&3 in conf_validation, but place the file in ./python/ to be used for the specific task of validating arguments for an ngen run. That way the classes used to validate the individual config files can be used directly from the ngen.conf library.

python/ngen_conf/tests/data/test_realization_config.json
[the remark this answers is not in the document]
JordanLaserGit: It is identical with the exception of the routing[' t_route_config_file_with_path'] which I had to set to the full path to keep from the realization validation from breaking. The path in python/ngen_conf/tests/test_config.json is a relative path within the test folder.  I missed a hack in test_cfe_sloth.py where the full path is created for that field. Will delete this fill and incorporate the fix into test_conf

JordanLaserGit:
> Thanks for making the requested changes @JordanLaserGit! As I re-reviewed this, I think that `geojson_pydantic` really simplified how this PR will accomplish it's goal. One thing that stuck out that would be nice to include here are pydantic models that capture the geojson geometry types and properties for the different hydrofabric geojson files. For example, a `nexus` file is a `FeatureCollection` of `Point` Features each with property `toid` that has a str value. In pydantic / geojson_pydantic that might look something like:
> 
> ```python
> from pydantic import BaseModel
> from geojson_pydantic import FeatureCollection, Feature, Point
> 
> 
> class NexusFeatureProperty(BaseModel):
>   toid: str
> 
>   # NOTE: you could add a validator here to check the `toid`'s prefix.
> 
> NexusFeatureCollection = FeatureCollection[Point, NexusFeatureProperty]
> ```
> 
> I think it makes sense that these live in their own module in `ngen.config` something like `hydrofabric.py`(better thoughts are always wanted!)?
> 
> One other thing to note, if we are going to depend on `geojson_pydantic`, that will need to be added to `ngen.config`'s `setup.cfg` so that `pip` will install it when `ngen.config` is installed.
> 
> One last aside, `pydantic` is great, but it also has a fair number of bells and whistles as im sure you are finding out. [Here](https://docs.pydantic.dev/usage/models/#model-properties) is a listing of methods and properties that each model gets. The ones I noticed you might be interested in are `.parse_file()` and `.parse_obj()`.

Thanks for the review again @aaraney. In the commits since, I have tried to address your suggestions (let me know if I missed anything) and also left a response to your comment on conf_validation. The recent commits reflect what I suggested doing there. Let me know what you think!

python/ngen_conf/src/ngen/config/conf_validation.py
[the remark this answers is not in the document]
aaraney: Thanks for adding some context! 

> Perhaps it makes sense then to keep steps 2&3 in conf_validation, but place the file in ./python/ to be used for the specific task of validating arguments for an ngen run. That way the classes used to validate the individual config files can be used directly from the ngen.conf library.

Yeah, I think that is great. We already have a few utility scripts in there that do related tasks. I think this will get us on a nice track. It wont be difficult to take the work you've added in `conf_validation.py` and turn it into a simple `argparse` cli tool.

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -0,0 +1,74 @@
...
+from geojson_pydantic import Feature, Point,  MultiPolygon
+
+def check_in_accepted(val, accepted):
+    if val not in accepted:
+        raise ValueError(f'Not an acceptable option. Options are {accepted}')
+    
+class NexusFeatureProperty(BaseModel):
+    toid: str
+
+    @validator('toid')
```
aaraney, line 13: I should have mentioned this earlier, so my apologies for not, the `toid` prefix (i.e. `cat-`) has not yet stabilized in the hydrofabric. The older hydrofabric's use `cat-`, but newer versions are using `wb-` and I forget, but I seen other prefixes too. I wish I has a good answer for why...

I think the best bet here is to use a regular expression to validate the `toid` field. Luckily, the all do follow the pattern `{{ some-prefix }}-{{ numbers }}`, so you should be able to use the regular expression `"-\d+$"`. In practice that will look like:

```python
import re

# match str that ends with a '-' followed by digits
_TOID_REGEX_STR = "-\d+$"
_TOID_REGEX = re.compile(_TOID_REGEX_STR)

class NexusFeatureProperty(BaseModel):
    toid: str

    @validator('toid')
    def _check_prefix(cls, value: str) -> str:
        if _TOID_REGEX.search(value) is None:
            raise ValueError(f"Invalid 'toid' property value: {value!r}")
        return value

```

The same applies `CatchmentFeatureProperty`, the `toid` prefix is now either `nex-` or `tnx-` (terminal nexus), but just to be on the safe side, I would use the regex for this too. So really you could reuse the validator if you want like:


```python
def _validate_prefix(value: str) -> str:
    if _TOID_REGEX.search(value) is None:
        raise ValueError(f"Invalid 'toid' property value: {value!r}")
    return value
```

Then in the models just do:

```python
class NexusFeatureProperty(BaseModel):
    toid: str

    _validate_prefix = validator("toid", allow_reuse=True)(_validate_prefix)
```

Sorry, that is a lot. There is a fair amount of head scratching when it comes to ngen configuration files.

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -0,0 +1,74 @@
...
+    toid: str
+    areasqkm : float
+
+    @validator('toid')
+    def check_prefix(cls,val):
+        assert val[:3] == 'nex', "toid must be a nexus id"
+
+CatchmentFeature = Feature[MultiPolygon, CatchmentFeatureProperty]
+
+class CRSProperty(BaseModel):
```
aaraney, line 29: This is [optional in the GeoJSON spec](https://gist.github.com/sgillies/1233327#file-geojson-spec-1-0-L113), so I think it is a little overkill and potentially error prone to validate this. I feel even better about not validating it given that [rfc7946](https://www.rfc-editor.org/rfc/rfc7946), the GeoJSON Format notes in 3.4: 

> The coordinate reference system for all GeoJSON coordinates is a
   geographic coordinate reference system, using the World Geodetic
   System 1984 (WGS 84) [[WGS84](https://www.rfc-editor.org/rfc/rfc7946#ref-WGS84)] datum, with longitude and latitude units
   of decimal degrees.  This is equivalent to the coordinate reference
   system identified by the Open Geospatial Consortium (OGC) URN
   urn:ogc:def:crs:OGC::CRS84

 I appreciate the thoroughness though :smile:!

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -0,0 +1,74 @@
...
+    """    
+    name: Literal["urn:ogc:def:crs:OGC:1.3:CRS84"]
+
+    # Are any others accepted? If so, check with validator
+    # @validator('name')
+    # def check_properties(cls,val):
+    #     accepted = ["urn:ogc:def:crs:OGC:1.3:CRS84"] # Are any others accepted?
+    #     check_in_accepted(val, accepted)
+
+class CRS(BaseModel):
```
aaraney, line 41: Likewise with the above comment, I think this model can go too.

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    type: Literal["name"]
+    properties: CRSProperty
+
+    @validator('type')
+    def check_type(cls,val):
+        accepted = ["name"]
+        check_in_accepted(val, accepted)
+
+class NGenCatchmentFile(BaseModel):
```
aaraney, line 53: I think we can simplify the two following models to just:

```python
CatchmentGeoJSON = FeatureCollection[MultiPolygon, CatchmentFeatureProperty]

NexusGeoJSON = FeatureCollection[Point, NexusFeatureProperty]
```

python/ngen_conf/tests/test_conf.py
```diff
@@ -0,0 +1,46 @@
+import pytest, json
+from pathlib import Path
+from ngen.config.realization import NgenRealization
+from ngen.config.hydrofabric import NGenNexusFile, NGenCatchmentFile
+
+@pytest.fixture()
+def catchmentdata():
```
aaraney, line 7: As an aside, just a cool little nugget about `pytest`, fixtures can take other fixtures as arguments. So you could have a `test_dir` fixture that the `catchmentdata`, `nexusdata`, and `realizationdata` fixtures each use used. You pass them in just like you would a fixture to a test. Its totally fine how it is, just trying to teach a little along the way!

python/conf_validation.py
```diff
@@ -0,0 +1,89 @@
+# Entry point to validating NGen catchment,nexus, and realization files
+import sys, json
+from ngen.config.realization import NgenRealization
+from ngen.config.hydrofabric import NGenCatchmentFile, NGenNexusFile
+
+def main(catchment_file,catchment_subset,nexus_file,nexus_subset,rel_file):
```
aaraney, line 6: I think in the future we will want to turn this into a little CLI tool that has documentation and what not, but for now I don't think it is worth the effort.

python/conf_validation.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    validate the three config files and sub selections
+    """
+    # Validate Catchment config
+    with open(catchment_file) as fp:
```
aaraney, line 11: For the most part, the rest of my comments can be ignored in this file. I wanted to offer some pointers (no pun intended) that might make your life a little easier and de-duplicate work.

You could re-write the catchment validation like:
```python
serialized_catchments = NGenCatchmentFile.parse_file(catchment_file)

catchments = []
catchment_pairs = []

for feature in serialized_catchments.features:
  id = feature.id
  toid = feature.properties.toid

  catchment_pairs.append((id, toid))
  catchments.append(id)

```

aaraney: You could do the same thing for `NGenNexusFile` too!

aaraney (changes requested):
Thanks for working through this with me, @JordanLaserGit! I left a few more, mostly minor, changes and then I think we will be good to merge!

python/ngen_conf/src/ngen/config/hydrofabric.py
JordanLaserGit: This all makes sense. I added the condition that the characters before the hyphen can't contain a number. I think it's possible to require one of several letter combinations ("cat" or "wb") if we want.

python/ngen_conf/src/ngen/config/hydrofabric.py
JordanLaserGit: Yup. Figured we may want to validate that field, but we can just leave it out if not.

python/ngen_conf/src/ngen/config/hydrofabric.py
JordanLaserGit: This makes sense if we don't want validate the name or crs fields. Sounds like we don't so, the recent push reflects this change.

python/ngen_conf/tests/test_conf.py
JordanLaserGit: Took your suggestion and made this change

python/conf_validation.py
JordanLaserGit: Took this suggestion too. Makes the code look cleaner and why not use a bell/whistle if we got it!

JordanLaserGit:
I implemented your suggestions, @aaraney. Thanks for the feed back and let me know if anything else needs adjusting.

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -1,74 +1,35 @@
...
 from pydantic import BaseModel, validator
-from geojson_pydantic import Feature, Point,  MultiPolygon
+from geojson_pydantic import FeatureCollection, Point,  MultiPolygon
+import re
 
-def check_in_accepted(val, accepted):
-    if val not in accepted:
-        raise ValueError(f'Not an acceptable option. Options are {accepted}')
+# match str that ends with a '-' followed by digits
+_TOID_REGEX_STR = ".\D+-\d+$"
```
aaraney, line 8: I think we want `"^\w+-\d+$"` here. `\D` captures any non-digit character, including `-`. `\w` will only capture word characters (including `_`).

python/ngen_conf/src/ngen/config/hydrofabric.py
```diff
@@ -1,74 +1,35 @@
...
-    Model for catchment data file
-    """
-
-    # required
-    type: Literal["FeatureCollection"]
-    name: Literal["catchment_data"]
-    crs: CRS
-    features: List[CatchmentFeature]
-
+    def _validate_toid(cls, value: str) -> str:
```
aaraney, line 16: Sorry, I think this was lost in communication. This function can be moved outside of `CatchmentFeatureProperty` scope. So:

```python

def _validate_toid(value: str) -> str:
    if _TOID_REGEX.search(value) is None:
        raise ValueError(f"Invalid 'toid' property value: {value!r}")
    return value

class CatchmentFeatureProperty(BaseModel):
    toid: str
    areasqkm : float

    _validate_toid = validator("toid", allow_reuse=True)(_validate_toid)


class NexusFeatureProperty(BaseModel):
    toid: str

    _validate_toid = validator("toid", allow_reuse=True)(_validate_toid)
```

aaraney (changes requested):
Few minor code cleanup things and then I think we will be on the road.

JordanLaserGit:
Implemented your suggestions.

aaraney (changes requested):

JordanLaserGit:
Addressed.

aaraney (dismissed):
Looks good to me, @JordanLaserGit! Thanks for iterating through this with me! I am going to have @hellkite500 give it a quick glance before merging, but we should be good to go. Thanks again!

@hellkite500, this is ready for review and should be a quick one. Please take a look when you get a chance.

JordanLaserGit:
Hey @aaraney, realized the way I was validating the catchment/nexus topology was not exactly what we wanted to be doing. We can add more to this validation later if we want (making sure all the catchments are actually connected hydrologically). I just have in here a quick check that all nexus in the catchment file are present in the nexus file. 

I also added an optional class variable to the properties field. Reason being is geopandas places the feature id within the property field, so it's useful to have this.

aaraney (changes requested):
Minor changes, mainly just need to move `conf_validation.py` out of the `ngen.config` package into `/python`. Your other changes look good though! Thanks, @jordanLaserGit!

JordanLaserGit:
Thanks for looking at this again. Addressed your comments. Sorry about moving conf_validation again. I was thinking about importing it into another package, but then realized that would create a dependency that we probably don't want.

aaraney (approved):
Thanks again for working through this with me, @JordanLaserGit! Thanks for thinking to add this feature and making it happen. This will be super useful to numerous parties.

PR #49, opened 2023-04-05
BMI model init config pydantic models

aaraney:
This PR adds pydantic models for currently supported NGEN BMI models (i.e. `CFE`). This work makes it possible to serialize, deserialize, validate, and programmatically access supported BMI model configuration files in a consistent manor.

This work relies on #43. So, it should not be merged until #43 has.

~I still have a few models to port over, so for now, I am going to leave this as a draft and gradually port models to this PR.~

## Additions

- pydantic model of [`CFE`](https://github.com/NOAA-OWP/cfe/)'s configuration file.
- "" [`PET`](https://github.com/NOAA-OWP/evapotranspiration/)'s configuration file.
- "" [`NoahOWP`](https://github.com/NOAA-OWP/noah-owp-modular/tree/30d0f53e8c14acc4ce74018e06ff7c9410ecc13c)'s configuration file.
- "" [`LSTM`](https://github.com/NOAA-OWP/lstm)'s configuration file.

## Todos

- CFE has rootzone-based actual evapotranspiration (AET) and ice-fraction based runoff options that were not super clear. I have marked TODOs to go back and add these options once I have a better grasp on these options. I dont think these TODOs should hold up this PR.
- modify LSTM config model once a better understanding of how LSTM is to be configured in practice. Based on the configs present in https://github.com/NOAA-OWP/lstm it is unclear when certain keys are required. It is safe to merge this and leave this for later if so we choose.
- add LSTM tests

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #48, opened 2023-03-29
Add PathPair and PathPairCollection types

aaraney:
Add new `PathPair` and `PathPairCollection` types that encapsulate a `pathlib.Path` and some inner `T` or collection of `T`'s. Beyond encapsulation, these types also provide APIs for flexibly writing, reading, serializing, deserializing to and from the inner `T` / `T`'s.

This functionality is desirable, among other use cases, when using `pydantic` and want to represent a field that de/serializes as a `pathlib.Path` but holds onto an inner deserialized `T` (this could be another pydantic model for example).

```python
from pydantic import BaseModel, validator
from ngen.config.path_pair import PathPair
from ngen.config.path_pair.common import pydantic_serializer, pydantic_deserializer
import tempfile
from typing import Union

class Inner(BaseModel):
    foo: int

class Outer(BaseModel):
    path: PathPair[Inner]

    @validator("path", pre=True)
    def _coerce_path(cls, value: Union[PathPair[Inner], str]) -> PathPair[Inner]:
        if isinstance(value, PathPair):
            return value
        value = PathPair(
            value,
            serializer=pydantic_serializer,
            deserializer=pydantic_deserializer(Inner),
        )
        try:
            # read and deserialize from path into Inner
            value.read()
        except:
            ...
        return value

def test_outer():
    with tempfile.NamedTemporaryFile() as file:
        path_to_file = Path(file.name)

        inner = Inner(foo=12)
        inner_path_pair = PathPair.with_object(
            inner,
            path=path_to_file,
            serializer=pydantic_serializer,
            deserializer=pydantic_deserializer(Inner),
        )
        # serialize inner T, Inner and **write to disc**
        assert inner_path_pair.write() == True

        # when pydanatic validates this, we will **read in and deserialize**
        # into the Inner type
        outer = Outer(path=path_to_file)
        assert outer.path == path_to_file
        assert outer.path.inner == inner
```

<details>
<summary><code>PathPairCollection</code> example</summary>

```python
class OuterWithPathPairCollection(BaseModel):
    path: PathPairCollection[Inner]

    _path_pattern: ClassVar[str] = "{id}"

    @validator("path", pre=True)
    def _coerce_path(
        cls, value: Union[PathPairCollection[Inner], str]
    ) -> PathPairCollection[Inner]:
        if isinstance(value, PathPairCollection):
            return value

        value = PathPairCollection(
            value,
            pattern=cls._path_pattern,
            serializer=pydantic_serializer,
[...]

PR #50, opened 2023-05-01
Realization validation

JordanLaserGit:
Implemented config and forcing path validation

## Additions
- config and forcing path validation consistent with AWI catchment file convention (cat{vpu}_{id}.csv).
- catchments to test_config.json

## Removals

- bmi_fortran module 

## Changes

-

## Testing

1. pytests for ngen_conf pass

## Notes

-

## Todos

- The bmi_fortran module should be included in test_config.json, need to figure out why this is failing the validation.

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

- [ ] Linux
- [ ] MacOS


aaraney:
@JordanLaserGit, thanks for opening this! I don't think that we want to strictly validate realization config files at instance creation time, as you have it here. I _do_ think we want the functionality to more stringently validate config files, however I think we want to do that as an optional postponed validation step. Specifically, we don't want to strictly enforce that referenced (e.g. `init_config`) files in a `NgenRealization` exist during instance creation. This is not desirable because it limits the contexts and situations in which we can use this tooling. For example, if we were programmatically creating a `NgenRealization` instance, it would be pre-requisite that we have the associated bmi module `init_config` files on disk before creation. Otherwise, `pydantic` validation would fail. So it is kind of a chicken and egg problem... There are other examples like configuration generation where this would not be desirable. You might also be interested in #48 and #49. These PR's are related to what you are working on here. 

So, what does that mean? Well, I think the best that we can do for now is provide this functionality in the `python/conf_validation.py` script. I am thinking a function that takes a pydantic model and then walks that model looking for fields that are `Path` instances and verifies that they exist. You should be able to use the magic property `__fields__` on pydantic `BaseModel` instances and `getattr` to achieve this. If we do move in that direction, it might be best if we close this PR and open a new PR on a fresh branch for history sake. Thoughts?

hellkite500:
The functionality in practice is quite useful, but needs to be decoupled a bit more from the backend config library.  We can definitely work to build the utility out in the future expanding on the work `conf_validation.py` in a more stand-alone fashion.

PR #58, opened 2023-07-18
Change minimum ngen.init_config python version >=3.7

aaraney:
`ngen.cal` and `ngen.config` both support `3.7`. In introducing `ngen.init_config` I overlooked that minimum version and set it to `>=3.8`. This PR makes the necessary changes to bring that down to `>=3.7` and changes the minimum supported version in the package metadata.

## Changes

- `ngen.init_config`'s minimum supported python version >=3.7 from >=3.8


aaraney:
This should fix #49's failing tests once it is merged.

hellkite500 (approved):

PR #59, opened 2023-07-18
Update ngen-init config gh actions

aaraney:
Update `ngen.init_config` gh actions to mirror our other actions.


hellkite500 (approved):
Should we update all the test matrices for 3.7-3.11…?

aaraney:
> Should we update all the test matrices for 3.7-3.11…?

@hellkite500, yeah. I would also like to print out the dep versions as a step too. I can get a PR up for that shortly.

PR #82, opened 2023-11-21
`to_{{format}}_str` methods always return non-`eol` terminated `str`; `to_{{format}}` methods always terminate file with `eol`

aaraney:
Previously, `ini`, `yaml`, and `toml` serializers returned `eol` terminated `str`s while `namelist` and `json` serializers returned non-`eol` terminated `str`s. This lead to some types writing an `eol` character to a file while others did not. This PR updates the behavior so that all `to_{{format}}_str` methods return `str`s that have _not_ been terminated with an `eol` and all `to_{{format}}` methods write `eol` characters.

## Changes

- `to_{{format}}_str` methods always return non-`eol` terminated `str`s.
- `to_{{format}}` methods always write a `eol` trailing character.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:


aaraney:
Alternatively, we could remove the `eol` character from all `str` methods and add it only when writing to a file.

aaraney:
The more I think about this, the more that I waver back and forth. I think the better thing to do is to alway return `str`s without an `eol` character and always write to files with an `eol` character. That is at least what I would expect as a user of a library.

aaraney:
The tests are failing b.c. `ngen_cal` is using the current HEAD version of `ngen.init_config` instead of the work has been introduced here.

PR #92, opened 2024-01-17
Proposal: Derive pkg version from git using `setuptool_scm`

aaraney:
Given that our packages are not published to a public registry (i.e. `PyPI`), anyone who uses these packages either installs using a github link via pip or by cloning the repo and installing from a local copy. When issues arise and user submit issues, a common first question is, "what version of the code are you using?" For worse, at this moment, we/users dont have a robust way of answering that question. As it sits, we hardcode the version of a package in a subpackage's `_version.py` file and call it a day. However, as we all know, this does not the the whole story unless we update `_version.py` with every commit to the repo (insanity). To remedy this, this PR introduces [`setuptool_scm`](https://setuptools-scm.readthedocs.io/en/latest/), a build time tool that extracts information from `git` and adds version information to packages including a `_version.py` file, as a build time step to all of our packages. This works when installing from a git link, locally, and locally in edit mode.

How does it work? `setuptool_scm` uses git tags to determine versions. It searches for the latest git tag and creates a version string based on the tag. See the [table of the default version scheme](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme) for more information.

Given that this repo is effectivity a monorepo, i've chosen to introduce a tagging convention that lets us track subpackage versions using tagging namespaces (i.e. `ngen.<package-name>/v<x>.<y>.<z>`). So for example, to cut a release of the `ngen.config` subpackage, just push an _annotated_ tag like `git tag --annotated "ngen.config/v0.2.5" --message "ngen.config v0.2.5"; git push upstream "ngen.config/v0.2.5"`.

To aide in the transition, ive created tags for all releases of all subpackages and pushed them to the repo. Even if we decide against this, this will be helpful.

related to https://github.com/NOAA-OWP/DMOD/issues/476

## Notes

- Tags follow the following semantics: `ngen.<package-name>/v<x>.<y>.<z>`. Tags that do not follow this format will not be recognized and used by `setuptools_scm` to create the correct pkg version.


aaraney:
Looks like `setuptools_scm` does not have a `>=8` version for 3.7. Ill look into if we can get around this tomorrow.

aaraney:
Moving this into draft. There are still a few things I need to sort out before this is ready for review.

PR #95, opened 2024-02-06
feat: `value_unit_pair` module

aaraney:
Add `ValueUnitPair` and `ListUnitPair` classes. The former captures the concept of a value and a unit (i.e. `42[meters]`) while the latter is an extension of the concept but for a list of values.

Why do we need this? BMI modules: soil freeze thaw, soil moisture profile, lgar, and CFE all use this format in their configuration files. These classes provide a way for capturing the expressed concept as pydantic models. This will enable adding the aforementioned modules to `ngen.config`.

## Notes

- There are a few PRs that I am in the process of staging that will result in a `ngen.config` version bump, but this alone does not constitute a version bump.

### DIFF
--- a/python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
+++ b/python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
@@ -1,0 +1,95 @@
+import re
+from typing import TYPE_CHECKING, Any, Generic, List, Optional, Type, TypeVar, Union
+
+from pydantic import validator
+from pydantic.generics import GenericModel
+from typing_extensions import Self, override
+
+if TYPE_CHECKING:
+    from pydantic.typing import AbstractSetIntStr, DictStrAny, MappingIntStrAny
+
+V = TypeVar("V")
+"""The type of `ValueUnitPair`'s `value` field."""
+U = TypeVar("U")
+"""The type of `ValueUnitPair`'s `unit` field."""
+
+
+_VAL_UNIT_RE = re.compile(r"(.*)\[(.*)\]")
+
+
+class ValueUnitPair(GenericModel, Generic[V, U]):
+    value: V
+    unit: U
+
+    @override
+    @classmethod
+    def validate(cls: Type[Self], value: Any) -> Self:
+        if isinstance(value, ValueUnitPair):
+            # return a shallow copy. this also validates / coerces mismatching generic types
+            return cls(value=value.value, unit=value.unit)
+
+        # unpack kwargs like arguments into expected string form
+        if isinstance(value, dict):
+            v = value.get("value", Ellipsis)
+            u = value.get("unit", Ellipsis)
+            if v == Ellipsis or u == Ellipsis:
+                raise ValueError(f"cannot coerce value='{value!r}' into {cls.__name__}")
+            return cls(value=v, unit=u)
+
+        # cannot further coerce / validate value
+        if not isinstance(value, str):
+            raise ValueError(f"cannot coerce value='{value!r}' into {cls.__name__}")
+
+        match = _VAL_UNIT_RE.search(value)
+
+        if match is None:
+            raise ValueError(f"no match in str: {value!r}")
+
+        # examples
+        #   2[m] -> ("2", "m")
+        #   1,2,3,4[m/m] -> ("1,2,3,4", "m/m")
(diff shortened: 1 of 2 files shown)


python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import TYPE_CHECKING, Any, Generic, List, Optional, Type, TypeVar, Union
+
+from pydantic import validator
+from pydantic.generics import GenericModel
+from typing_extensions import Self, override
+
+if TYPE_CHECKING:
+    from pydantic.typing import AbstractSetIntStr, DictStrAny, MappingIntStrAny
+
+T = TypeVar("T")
```
hellkite500, line 11: I, for one, would appreciate some context around these type vars -- what are they abstractly representing in their usage below?  A good comment string could be helpful to bring that context to the module.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
```diff
@@ -0,0 +1,79 @@
...
+V = TypeVar("V")
+U = TypeVar("U")
+
+
+_VAL_UNIT_RE = re.compile(r"(.*)\[(.*)\]")
```
hellkite500, line 16: Not that I think we should now, but may want to consider more restrictive regex here in the future...especially considering captures which have spaces, non-standard characters, ect.  This may be impossible to generalize and too far out of scope though -- just a passing thought.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
```diff
@@ -0,0 +1,79 @@
...
+    @override
+    @classmethod
+    def validate(cls: Type[Self], value: Any) -> Self:
+        if isinstance(value, ValueUnitPair):
+            return value
+
+        if isinstance(value, dict):
+            value = f"{value['value']}[{value['unit']}]"
+
+        if not isinstance(value, str):
```
hellkite500, line 32: this extra type check here seems irrelevant and kinda confusing...should the previous `if` not simply return value after creating the format string, eliminating this check?

hellkite500 (changes requested):

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
aaraney: Yeah thanks for pointing this out. I will add a comment here. There is a lot of pydantic magic happening in this method that if you don't know _how_ pydantic works under the hook then you would be pretty understandably lost...

`validate` is special to pydantic. Its what actually validates and coerces some value into a type _or partially_ does that.  pydantic will iterate over the functions its gets from that `__get_validators__` dundermethod, passing the nth result to the nth+1 validator and so on. These are functions are shaped like the `validate` class method above. So, in this case, if at this point we don't have a string that we can do something with, we just return it and pydantic will pass it to the next validator method in the chain.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
aaraney: Yeah I agree. I didn't and don't know at this point in time what appropriate constraints to enforce here, so I left it a little open. My thinking was we could always refine this if needed.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
aaraney: Yeah that is fair. Ill add a docstring explaining their purpose a bit.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
hellkite500: But under what circumstance can we get to this point without having a string we can do something with?

aaraney: Yeah now that I look at it closer and poke around with it more, I think you are right. Great catch. There are several other edge cases that ive glossed over. I am addressing those and will request a re-review.

hellkite500 (approved):

aaraney:
Thanks, @hellkite500!

PR #118, opened 2024-03-26
feat: warn if soil or veg type is water but not both

aaraney:
Warn if noah owp modular's soil or veg type is water but not both.

related: 
- https://github.com/NOAA-OWP/DMOD/issues/472#issuecomment-1893759701 
- https://github.com/NOAA-OWP/hydrofabric/issues/35

## Changes

- Warn if noah owp modular's `isltyp` or `vegtyp` is a water classification but not both.


python/ngen_conf/src/ngen/config/init_config/noahowp.py
```diff
@@ -270,4 +271,27 @@ class Config(core.Base.Config):
...
+    if parameters.veg_class_name == "USGS":
+        VEG_USGS_WATER = 16
+        veg_type_is_water = structure.vegtyp == VEG_USGS_WATER
+    elif parameters.veg_class_name == "MODIFIED_IGBP_MODIS_NOAH":
+        VEG_MODIS_WATER = 17
+        veg_type_is_water = structure.vegtyp == VEG_MODIS_WATER
+    else:
+        return
+
+    if (veg_type_is_water) ^ (soil_type_is_water):
```
hellkite500, line 291: This only works "correctly" when the operands are `bool` types, which by the code above ensures they are, so you the logical and bitwise exclusive or are equivalent.  It can be a bit of a foot gun if this function is modified, or copied/used in the future for different validation without understanding this, but I don't see much need to change it now.  Perhaps a quick comment to the above and the use of bitwise XOR to save someone in the future some possible headache?

hellkite500 (dismissed):

python/ngen_conf/src/ngen/config/init_config/noahowp.py
aaraney: Yeah, I think that is fair. Thanks for mentioning it! Just added a comment.

hellkite500 (approved):

PR #125, opened 2024-05-17
fix: use correct _outflow variable (fixes #124)

### LINKED ISSUE #124: `outflow` instance parameter doesnt exist
Fairly certain this is a bug. I don't think the `outflow` instance parameter / property exists anywhere on the class hierarchy.

https://github.com/NOAA-OWP/ngen-cal/blob/b71bfb8551760843e382e9df53063c9e4701d220/python/ngen_cal/src/ngen/cal/calibration_cathment.py#L73

hellkite500:
Fix for #124


aaraney (approved):
looks good to me!

PR #128, opened 2024-06-21
Gpkg support

hellkite500:
Support geopackage hydrofabric inputs.  This maintains compatibility with the old geo-json input for now, but will print a deprecation warning when trying to use them.

Note that all hydrofabric keys are now optional in the config, but either hydrofabric or the the 3 geoson need to be provided for validation to succeed.  If both hydrofabric and geojson are provided, the geojson will be silently ignored (other than the deprecation warning...) and the hydrofabric input key will be used to initialize the model.

### DIFF
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -110,7 +110,7 @@ class NgenBase(ModelExec):
         shutil.copy(self.realization, str(self.realization)+'_original')
        
         #Read the catchment hydrofabric data
-        if(self.hydrofabric is not None):
+        if self.hydrofabric is not None:
             #Reading hydofabric from geopackage
             self._catchment_hydro_fabric = gpd.read_file(self.hydrofabric, layer='divides')
             self._catchment_hydro_fabric.set_index('divide_id', inplace=True)
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -6,6 +6,7 @@ except ImportError:
     from typing_extensions import Literal
 from pathlib import Path
 import logging
+import warnings
 #supress geopandas debug logs
 logging.disable(logging.DEBUG)
 import json
@@ -264,7 +265,7 @@ class NgenBase(ModelExec):
             except:
                 raise TypeError("hydrofabric must be a valid file path")
         if cats is not None or nex is not None or x is not None:
-            print("WARNING: GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.")
+            warnings.warn("GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.")
 
         return values
 
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -230,6 +230,37 @@ class NgenBase(ModelExec):
             raise ValueError("Must provide partitions if using parallel")
         return values
 
+    @root_validator(pre=True)
+    def validate_hydrofabic(cls, values: dict) -> dict:
+        """
+
+        Args:
+            values (dict): configuation values to check
+
+        Raises:
+            ValueError: If a geopackage hydrofabric or set of geojsons are not found
+
+        Returns:
+            dict: Valid configuration elements for hydrofabric requirements
+        """
+        hf: FilePath = values.get('hydrofabric')
+        cats: FilePath = values.get("catchments")
+        nex: FilePath = values.get("nexus")
+        x: FilePath = values.get("crosswalk")
+
+        if hf is None and cats is None and nex is None and x is None:
+            msg = "Must provide a geopackage input with the hydrofabric key"\
+                  "or proide catchment, nexus, and crosswalk geojson files."  
+            raise ValueError(msg)
+        if(hf is not None):
+            try:
+                p = Path(hf)
+                if not p.exists():
+                    raise
+            except:
+                raise TypeError("hydrofabric must be a valid file path")
+        return values
+
     def update_config(self, i: int, params: 'pd.DataFrame', id: str = None, path=Path("./")):
         """_summary_
 
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -256,7 +256,7 @@ class NgenBase(ModelExec):
                   "or proide catchment, nexus, and crosswalk geojson files."  
             raise ValueError(msg)
         
-        if(hf is not None):
+        if hf is not None:
             try:
                 p = Path(hf)
                 if not p.exists():
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -234,9 +234,11 @@ class NgenBase(ModelExec):
             raise ValueError("Must provide partitions if using parallel")
         return values
 
-    @root_validator(pre=True)
+    @root_validator()
     def validate_hydrofabic(cls, values: dict) -> dict:
         """
+        Validates hydrofabric information is provided either as (deprecated) GeoJSON
+        or (preferred) GeoPackage files.
 
         Args:
             values (dict): configuation values to check
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -252,6 +252,7 @@ class NgenBase(ModelExec):
             msg = "Must provide a geopackage input with the hydrofabric key"\
                   "or proide catchment, nexus, and crosswalk geojson files."  
             raise ValueError(msg)
+        
         if(hf is not None):
             try:
                 p = Path(hf)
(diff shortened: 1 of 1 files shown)


python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -74,9 +74,10 @@ class NgenBase(ModelExec):
     # an existing realization to build various calibration realization configs
     # but we should probably take a closer look at this in the near future
     realization: FilePath
-    catchments: FilePath
-    nexus: FilePath
-    crosswalk: FilePath
+    hydrofabric: Optional[FilePath]
```
aaraney, line 77: I didn't realize that these were using `pydantic.FilePath`. We really shouldn't be using that anywhere directly. During validation `FilePath` checks if the file exists and that is not always desirable (e.g. validating but don't care, yet, if the files exist). I wrote a [wrapper type](https://github.com/NOAA-OWP/t-route/blob/1a5ccf243db939c186c0cdfe00c080b25cce5ab0/src/troute-config/troute/config/types.py#L15) for `t-route` that kind of gets you the best of both worlds and allows you to turn on a strict mode when you do want to check that files are present during validation. This seems out of scope for this PR though. For now, can you change these to just `Optional[Path]` and leave a note that they should be file paths?

Ill open an issue to switch these over to something like what I introduced in `t-route`.

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -212,6 +230,41 @@ def check_for_partitions(cls, values: dict):
             raise ValueError("Must provide partitions if using parallel")
         return values
 
+    @root_validator(pre=True)
```
aaraney, line 233: I don't think this needs to be `pre` does it?

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -108,23 +110,35 @@ def __init__(self, **kwargs):
...
-        with open(self.crosswalk) as fp:
-            data = json.load(fp)
-            for id, values in data.items():
-                gage = values.get('Gage_no')
-                if gage:
-                    if not isinstance(gage, str):
-                        gage = gage[0]
-                    if gage != "":
-                        self._x_walk[id] = gage
+        if(self.hydrofabric is not None):
```
aaraney, line 113: Your c++ is showing 😏 
```suggestion
        if self.hydrofabric is not None:
```

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -108,23 +110,35 @@ def __init__(self, **kwargs):
...
+            self._catchment_hydro_fabric = gpd.read_file(self.hydrofabric, layer='divides')
+            self._catchment_hydro_fabric.set_index('divide_id', inplace=True)
+            self._nexus_hydro_fabric = gpd.read_file(self.hydrofabric, layer='nexus')
+            self._nexus_hydro_fabric.set_index('id', inplace=True)
+            self._flowpath_hydro_fabric = gpd.read_file(self.hydrofabric, layer='flowpaths')
+            self._flowpath_hydro_fabric.set_index('id', inplace=True)
+            attributes = gpd.read_file(self.hydrofabric, layer="flowpath_attributes").set_index('id')
+            self._x_walk = pd.Series( attributes[ ~ attributes['rl_gages'].isna() ]['rl_gages'] )
+        else:
+            #Reading hydrofabric from legacy geosjon
```
aaraney, line 124: Lets add a few `assert`s here to enforce the implicit invariants.

```suggestion
            # Legacy geojson support
            assert self.catchments is not None, "missing geojson catchments file"      
            assert self.nexus is not None, "missing geojson nexus file" 
            assert self.crosswalk is not None, "missing crosswalk file"
```

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -212,6 +230,41 @@ def check_for_partitions(cls, values: dict):
...
+        cats: FilePath = values.get("catchments")
+        nex: FilePath = values.get("nexus")
+        x: FilePath = values.get("crosswalk")
+
+        if hf is None and cats is None and nex is None and x is None:
+            msg = "Must provide a geopackage input with the hydrofabric key"\
+                  "or proide catchment, nexus, and crosswalk geojson files."  
+            raise ValueError(msg)
+        
+        if(hf is not None):
```
aaraney, line 256: :smile:
```suggestion
        if hf is not None:
```

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -212,6 +230,41 @@ def check_for_partitions(cls, values: dict):
...
+            try:
+                p = Path(hf)
+                if not p.exists():
+                    raise
+            except:
+                raise TypeError("hydrofabric must be a valid file path")
+        if cats is not None or nex is not None or x is not None:
+            print("WARNING: GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.")
```
aaraney, line 264: Lets use `warnings.warn` here instead.

aaraney (changes requested):
Overall this looks solid! There are a few stylistic changes I think are worth addressing plus a few minor necessary changes. Thanks, @hellkite500!

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: I wanted as early stop as possible, without hydro fabric the rest is kinda meaningless (at least for doing an actual calibration...).  It could probably be done effectively at any level...I was also just following in kind with the previous root validator...

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: For calibration config, I chose `FilePath` particularly for those semantics, as there wasn't much use case for calibration configs if they weren't being used to execute calibration, which makes these files paths required to exist for that to succeed.  I can be convinced that a use case exists which a calibration config would need to be validated "out of band" in which case adopting your wrapper proposal would help support both cases.

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: Yeah, that...I was trying to think of it at the time, and drew a huge blank 🤣

python/ngen_cal/src/ngen/cal/ngen.py
aaraney: It looks like you can just remove the `pre=True` parameter to accomplish the same thing. With `pre=True` the validator will be the first to handle the input and it seems like what you really want is that `pydantic` does its coercion / validation, then you apply these invariants. The `root_validator` will still run during object initialization, just after `pydantic` has done coercion / validation.

python/ngen_cal/src/ngen/cal/ngen.py
aaraney: 😆!

python/ngen_cal/src/ngen/cal/ngen.py
robertbartel: > ... there wasn't much use case for calibration configs if they weren't being used to execute calibration ...

At minimum, there is the use case for DMOD to create ngen-cal configurations at calibration job request time from user input (and there probably are others).  Using `FilePath` like this would effectively keep us from doing that without having such things all in the identical paths on whatever was creating the ngen-cal config, or doing some kind of lower-level string replacement on the resulting config file before it was used inside a worker.

hellkite500: Sounds reasonable, we will change these in another PR following #131

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: changed in [4db27d0](https://github.com/NOAA-OWP/ngen-cal/pull/128/commits/4db27d0b2a9b3fd66adad0db26f1079b0e3dd14a)

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -253,15 +257,15 @@ def validate_hydrofabic(cls, values: dict) -> dict:
...
+        if hf is not None:
             try:
                 p = Path(hf)
                 if not p.exists():
                     raise
             except:
                 raise TypeError("hydrofabric must be a valid file path")
         if cats is not None or nex is not None or x is not None:
-            print("WARNING: GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.")
+            warnings.warn("GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.")
```
aaraney, line 268:
```suggestion
            warnings.warn("GeoJSON support will be deprecated in a future release, use geopackage hydrofabric.", DeprecationWarning)
```

aaraney (changes requested):
Minor improvement to the warning then we should be good to go! Thanks for making the changes!

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: addressed in [8d60cb2](https://github.com/NOAA-OWP/ngen-cal/pull/128/commits/8d60cb23b31591671ebff9e3f5125435cc240aa6)

aaraney (approved):
Thanks, @hellkite500!

PR #129, opened 2024-06-21
User configurable evaluation feature

hellkite500:
Add configuration key `eval_feature` which allows a user to explicitly select which observable nexus/waterbody feature they want to use in the evaluation phase.  Previous behavoir was to raise an exception if the crosswalk had more than one gaged location.  Now it will throw this exception only if the `eval_feature` isn't set.  Single gage crosswalks will still work without this configuation key.

> [!NOTE]  
> commit based off #128, recommend merging after that PR.

### DIFF
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -76,6 +76,7 @@ class NgenBase(ModelExec):
     # but we should probably take a closer look at this in the near future
     realization: FilePath
     hydrofabric: Optional[FilePath]
+    eval_feature: Optional[str]
     catchments: Optional[FilePath]
     nexus: Optional[FilePath]
     crosswalk: Optional[FilePath]
@@ -420,9 +421,17 @@ class NgenIndependent(NgenBase):
             #FIXME pick up params per catchmment somehow???
             params = _map_params_to_realization(self.params, catchment)
             catchments.append(AdjustableCatchment(self.workdir, id, nexus, params))
+        
+        if self.eval_feature:
+            for n in eval_nexus:
+                wb = self._flowpath_hydro_fabric[ self._flowpath_hydro_fabric['toid'] == n.id ]
+                key = wb.iloc[0].name
+                if key == self.eval_feature:
+                    eval_nexus = [n]
+                    break
 
         if len(eval_nexus) != 1:
-            raise RuntimeError( "Currently only a single nexus in the hydrfabric can be gaged")     
+            raise RuntimeError( "Currently only a single nexus in the hydrfabric can be gaged, set the eval_feature key to pick one.")     
         self._catchments.append(CalibrationSet(catchments, eval_nexus[0], self.routing_output, start_t, end_t, self.eval_params))
 
     def _strip_global_params(self) -> None:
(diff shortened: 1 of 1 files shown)


python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -361,9 +415,17 @@ def __init__(self, **kwargs):
             #FIXME pick up params per catchmment somehow???
             params = _map_params_to_realization(self.params, catchment)
             catchments.append(AdjustableCatchment(self.workdir, id, nexus, params))
+        
+        if self.eval_feature:
+            for n in eval_nexus:
+                wb = self._flowpath_hydro_fabric[ self._flowpath_hydro_fabric['toid'] == n.id ]
```
aaraney, line 421: Little code golf for you.
```suggestion
                key = self._flowpath_hydro_fabric.loc[ self._flowpath_hydro_fabric['toid'] == n.id, 'name'].head(1)
```

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -408,8 +470,18 @@ def __init__(self, **kwargs):
             location = NWISLocation(nwis, nexus_data.name, nexus_data.geometry)
             nexus = Nexus(nexus_data.name, location, (), id)
             eval_nexus.append( nexus )
+        
+        if self.eval_feature:
+            for n in eval_nexus:
+                wb = self._flowpath_hydro_fabric[ self._flowpath_hydro_fabric['toid'] == n.id ]
+                print(wb)
```
aaraney, line 477: Seems like this was left in by mistake. Assume you were print debugging.

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -408,8 +470,18 @@ def __init__(self, **kwargs):
             location = NWISLocation(nwis, nexus_data.name, nexus_data.geometry)
             nexus = Nexus(nexus_data.name, location, (), id)
             eval_nexus.append( nexus )
+        
+        if self.eval_feature:
+            for n in eval_nexus:
+                wb = self._flowpath_hydro_fabric[ self._flowpath_hydro_fabric['toid'] == n.id ]
```
aaraney, line 476: Could do the same code golf I did above here.

aaraney (approved):
Left a few comments that are more or less code golf, but ill let you be the judge.

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: This is the hazards of complex indexing -- it can be written many ways, and miss understood in all of them haha.  I'm in favor of scoring this golf round by readability and complexity. 

Also, I'm not convinced that this will actually work, because `name` is *not* a column in the dataframe -- the `name` attribute used becomes essentially the index value of the returned series from `iloc`.

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: Removed and rebased along with a branch update from gpkg-support

aaraney (approved):

PR #130, opened 2024-06-21
Update to use t-route output in csv form

hellkite500:
At some point t-route stopped supporting `flowveldepth` output in hdf5 form.  However, it will still output the `flowveldepth` in a csv file which can be used very similarly.  This PR updates routing output to support reading the csv output instead of hdf5.

Also, when multiple runs of t-route are used, a cleanup of the `*NEXOUT.parquet` files is needed, so that was added here in the `update_config` step to ensure a clean directory for each calibration iteration.

Only works once noaa-owp/t-route#788 is applied, otherwise t-route will still throw an error when run through ngen-cal since the working directory contains other, non-troute specific, parquet files (see noaa-owp/t-route#787)

### DIFF
--- a/python/ngen_cal/src/ngen/cal/_hookspec.py
+++ b/python/ngen_cal/src/ngen/cal/_hookspec.py
@@ -8,6 +8,8 @@ from ngen.cal import PROJECT_SLUG
 
 if TYPE_CHECKING:
     from ngen.cal.configuration import General
+    from pandas import Series
+    from pathlib import Path
 
 hookspec = pluggy.HookspecMarker(PROJECT_SLUG)
 
@@ -42,3 +44,21 @@ def ngen_cal_finish(exception: Exception | None) -> None:
     raised during the calibration loop.
     `exception` will be non-none if an exception was raised during calibration.
     """
+
+class ModelHooks():
+    @hookspec(firstresult=True)
+    def ngen_cal_model_output(id: str | None) -> Series:
+        """
+            Called during each calibration iteration to provide the model output in the form
+            of a pandas Series, indexed by time.
+            Output series should be in units of cubic meters per second.
+        """
+
+    @hookspec
+    def ngen_cal_model_post_iteration(path: Path, iteration: int) -> None:
+        """
+            Called after each model iteration is completed and evaluated.
+            And before the next iteration is configured and started.
+            Currently called at the end of an Adjustable's check_point function
+            which writes out calibration/parameter state data each iteration.
+        """
--- a/python/ngen_cal/src/ngen/cal/_hookspec.py
+++ b/python/ngen_cal/src/ngen/cal/_hookspec.py
@@ -56,7 +56,7 @@ class ModelHooks:
         """
 
     @hookspec
-    def ngen_cal_model_iteration_finish(self, info: JobMeta, iteration: int) -> None:
+    def ngen_cal_model_iteration_finish(self, iteration: int, info: JobMeta) -> None:
         """
             Called after each model iteration is completed and evaluated.
             And before the next iteration is configured and started.
--- a/python/ngen_cal/src/ngen/cal/_hookspec.py
+++ b/python/ngen_cal/src/ngen/cal/_hookspec.py
@@ -46,7 +46,7 @@ def ngen_cal_finish(exception: Exception | None) -> None:
     `exception` will be non-none if an exception was raised during calibration.
     """
 
-class ModelHooks():
+class ModelHooks:
     @hookspec(firstresult=True)
     def ngen_cal_model_output(self, id: str | None) -> Series:
         """
--- a/python/ngen_cal/src/ngen/cal/_hookspec.py
+++ b/python/ngen_cal/src/ngen/cal/_hookspec.py
@@ -8,6 +8,7 @@ from ngen.cal import PROJECT_SLUG
 
 if TYPE_CHECKING:
     from ngen.cal.configuration import General
+    from ngen.cal.meta import JobMeta
     from pandas import Series
     from pathlib import Path
 
@@ -47,7 +48,7 @@ def ngen_cal_finish(exception: Exception | None) -> None:
 
 class ModelHooks():
     @hookspec(firstresult=True)
-    def ngen_cal_model_output(id: str | None) -> Series:
+    def ngen_cal_model_output(self, id: str | None) -> Series:
         """
             Called during each calibration iteration to provide the model output in the form
             of a pandas Series, indexed by time.
@@ -55,7 +56,7 @@ class ModelHooks():
         """
 
     @hookspec
-    def ngen_cal_model_post_iteration(path: Path, iteration: int) -> None:
+    def ngen_cal_model_iteration_finish(self, info: JobMeta, iteration: int) -> None:
         """
             Called after each model iteration is completed and evaluated.
             And before the next iteration is configured and started.
--- a/python/ngen_cal/src/ngen/cal/calibratable.py
+++ b/python/ngen_cal/src/ngen/cal/calibratable.py
@@ -92,13 +92,6 @@ class Adjustable(ABC):
         """
         self._df = read_parquet(path/self.check_point_file)
 
-    @abstractmethod
-    def save_output(self, i: int) -> None:
-        """
-            Save the last output of the runtime for iteration i
-        """
-        pass
-
     def restart(self) -> None:
             self.load_df('./')
 
--- a/python/ngen_cal/src/ngen/cal/calibratable.py
+++ b/python/ngen_cal/src/ngen/cal/calibratable.py
@@ -83,7 +83,7 @@ class Adjustable(ABC):
         """
         return Path('{}_parameter_df_state.parquet'.format(self.id))
 
-    def check_point(self, info: JobMeta, iteration: int) -> None:
+    def check_point(self, iteration: int, info: JobMeta) -> None:
         """
             Save calibration information
         """
--- a/python/ngen_cal/src/ngen/cal/calibratable.py
+++ b/python/ngen_cal/src/ngen/cal/calibratable.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from pandas import Series, read_parquet # type: ignore
 from typing import Optional, TYPE_CHECKING
@@ -8,7 +10,8 @@ if TYPE_CHECKING:
     from pathlib import Path
     from datetime import datetime
     from typing import Tuple, Callable, Optional
-    from .model import EvaluationOptions
+    from ngen.cal.model import EvaluationOptions
+    from ngen.cal.meta import JobMeta
 
 class Adjustable(ABC):
     """
@@ -80,10 +83,11 @@ class Adjustable(ABC):
         """
         return Path('{}_parameter_df_state.parquet'.format(self.id))
 
-    def check_point(self, path: 'Path', iteration: int) -> None:
+    def check_point(self, info: JobMeta, iteration: int) -> None:
         """
             Save calibration information
         """
+        path = info.workdir
         self.df.to_parquet(path/self.check_point_file)
 
     def load_df(self, path: 'Path') -> None:
--- a/python/ngen_cal/src/ngen/cal/calibratable.py
+++ b/python/ngen_cal/src/ngen/cal/calibratable.py
@@ -80,7 +80,7 @@ class Adjustable(ABC):
         """
         return Path('{}_parameter_df_state.parquet'.format(self.id))
 
-    def check_point(self, path: 'Path') -> None:
+    def check_point(self, path: 'Path', iteration: int) -> None:
         """
             Save calibration information
         """
--- a/python/ngen_cal/src/ngen/cal/calibration_set.py
+++ b/python/ngen_cal/src/ngen/cal/calibration_set.py
@@ -5,11 +5,13 @@ import pandas as pd
 if TYPE_CHECKING:
     from pandas import DataFrame
     from pathlib import Path
+    from pluggy import HookRelay
     from datetime import datetime
     from typing import Tuple, Optional
     from .model import EvaluationOptions
 import os
 from pathlib import Path
+import warnings
 from hypy.nexus import Nexus
 from .calibratable import Adjustable, Evaluatable
 
@@ -19,14 +21,15 @@ class CalibrationSet(Evaluatable):
         A HY_Features based catchment with additional calibration information/functionality
     """
 
-    def __init__(self, adjustables: Sequence[Adjustable], eval_nexus: Nexus, routing_output: 'Path', start_time: str, end_time: str, eval_params: 'EvaluationOptions'):
+    def __init__(self, adjustables: Sequence[Adjustable], eval_nexus: Nexus, hooks: 'HookRelay', start_time: str, end_time: str, eval_params: 'EvaluationOptions'):
         """
 
         """
         super().__init__(eval_params)
         self._eval_nexus = eval_nexus
         self._adjustables = adjustables
-        self._output_file = routing_output
+        # record the hooks needed for output
+        self._output_hook = hooks.ngen_cal_model_output
 
         #use the nwis location to get observation data
         obs =self._eval_nexus._hydro_location.get_data(start_time, end_time)
@@ -53,41 +56,22 @@ class CalibrationSet(Evaluatable):
             This re-reads the output file each call, as the output for given calibration catchment changes
             for each calibration iteration. If it doesn't exist, should return None
         """
-        try:
-            #in this case, look for routed data
-            #this is really model specific, so not as generalizable the way this
-            #is coded right now =(
-            #would be better to hook this from the model object???
-            #read the routed flow at the eval_nexus
-            df = pd.read_csv(self._output_file, index_col=0)
-            df.index = df.index.map(lambda x: 'wb-'+str(x))
-            tuples = [ eval(x) for x in df.columns ]
-            df.columns = pd.MultiIndex.from_tuples(tuples)
-            # TODO should contributing_catchments be singular??? assuming it is for now...
-            df = df.loc[self._eval_nexus.contributing_catchments[0].replace('cat', 'wb')]
-            self._output = df.xs('q', level=1, drop_level=False)
-            #This is a hacky way to get the time index...pass the time around???
-            tnx_file = list(Path(self._output_file).parent.glob("nex*.csv"))[0]
-            tnx_df = pd.read_csv(tnx_file, index_col=0, parse_dates=[1], names=['ts', 'time', 'Q']).set_index('time')
-            dt_range = pd.date_range(tnx_df.index[0], tnx_df.index[-1], len(self._output.index)).round('min')
-            self._output.index = dt_range
-            #this may not be strictly nessicary...I think the _evalutate will align these...
-            self._output = self._output.resample('1H').first()
-            self._output.name="sim_flow"
-            # self._output = read_csv(self._output_file, usecols=["Time", self._output_var], parse_dates=['Time'], index_col='Time', dtype={self._output_var: 'float64'})
-            # self._output.rename(columns={self._output_var:'sim_flow'}, inplace=True)
-            hydrograph = self._output
+        # TODO should contributing_catchments be singular??? assuming it is for now...
+        # Call output hooks, take first non-none result provided from hooks (called in LIFO order of registration)
+        df = self._output_hook(id=self._eval_nexus.contributing_catchments[0].replace('cat', 'wb'))
+        if not df:
+            # list of results is empty
+            print("No suitable output found from output hooks...")
+            df = None
+        elif len(df) > 1:
+            # TODO with the hook being firstresult=True, I don't think this is possible to hit...
+            warnings.warn("Multiple output data found, using first registered")
+            df = df[0]
+        else:
+            df = df[0]
+        return df
 
-        except FileNotFoundError:
-            print("{} not found. Current working directory is {}".format(self._output_file, os.getcwd()))
-            print("Setting output to None")
-            hydrograph = None
-        except Exception as e:
-            raise(e)
-        #if hydrograph is None:
-        #    raise(RuntimeError("Error reading output: {}".format(self._output_file)))
-        return hydrograph
-
+    # TODO should we still allow a setter here given the output hook used for this property?
     @output.setter
     def output(self, df):
         self._output = df
@@ -134,11 +118,11 @@ class UniformCalibrationSet(CalibrationSet, Adjustable):
         A HY_Features based catchment with additional calibration information/functionality
     """
 
-    def __init__(self, eval_nexus: Nexus, routing_output: 'Path', start_time: str, end_time: str, eval_params: 'EvaluationOptions', params: dict = {}):
+    def __init__(self, eval_nexus: Nexus, hooks: 'HookRelay', start_time: str, end_time: str, eval_params: 'EvaluationOptions', params: dict = {}):
         """
 
         """
-        super().__init__(adjustables=[self], eval_nexus=eval_nexus, routing_output=routing_output, start_time=start_time, end_time=end_time, eval_params=eval_params)
+        super().__init__(adjustables=[self], eval_nexus=eval_nexus, hooks=hooks, start_time=start_time, end_time=end_time, eval_params=eval_params)
         Adjustable.__init__(self=self, df=DataFrame(params).rename(columns={'init': '0'}))
 
         #For now, set this to None so meta update does the right thing
--- a/python/ngen_cal/src/ngen/cal/calibration_set.py
+++ b/python/ngen_cal/src/ngen/cal/calibration_set.py
@@ -59,10 +59,10 @@ class CalibrationSet(Evaluatable):
             #is coded right now =(
             #would be better to hook this from the model object???
             #read the routed flow at the eval_nexus
-            df = pd.read_hdf(self._output_file)
+            df = pd.read_csv(self._output_file, index_col=0)
             df.index = df.index.map(lambda x: 'wb-'+str(x))
-            df.columns = pd.MultiIndex.from_tuples(df.columns)
-
+            tuples = [ eval(x) for x in df.columns ]
+            df.columns = pd.MultiIndex.from_tuples(tuples)
             # TODO should contributing_catchments be singular??? assuming it is for now...
             df = df.loc[self._eval_nexus.contributing_catchments[0].replace('cat', 'wb')]
             self._output = df.xs('q', level=1, drop_level=False)
--- a/python/ngen_cal/src/ngen/cal/calibration_set.py
+++ b/python/ngen_cal/src/ngen/cal/calibration_set.py
@@ -33,8 +33,7 @@ class CalibrationSet(Evaluatable):
         self._eval_nexus = eval_nexus
         self._adjustables = adjustables
         # record the hooks needed for output and checkpointing
-        self._output_hook = hooks.ngen_cal_model_output
-        self._iteration_finish_hook = hooks.ngen_cal_model_iteration_finish
+        self._hooks = hooks
 
         #use the nwis location to get observation data
         obs =self._eval_nexus._hydro_location.get_data(start_time, end_time)
@@ -63,7 +62,7 @@ class CalibrationSet(Evaluatable):
         """
         # TODO should contributing_catchments be singular??? assuming it is for now...
         # Call output hooks, take first non-none result provided from hooks (called in LIFO order of registration)
-        df = self._output_hook(id=self._eval_nexus.contributing_catchments[0].replace('cat', 'wb'))
+        df = self._hooks.ngen_cal_model_output(id=self._eval_nexus.contributing_catchments[0].replace('cat', 'wb'))
         if not df:
             # list of results is empty
             print("No suitable output found from output hooks...")
@@ -102,7 +101,7 @@ class CalibrationSet(Evaluatable):
         for adjustable in self.adjustables:
             adjustable.df.to_parquet(info.workdir/adjustable.check_point_file)
         # call any model post hooks
-        self._iteration_finish_hook(info = info, iteration = iteration)
+        self._hooks.ngen_cal_model_iteration_finish(info = info, iteration = iteration)
 
     def restart(self) -> int:
         try:
--- a/python/ngen_cal/src/ngen/cal/calibration_set.py
+++ b/python/ngen_cal/src/ngen/cal/calibration_set.py
@@ -137,14 +137,6 @@ class UniformCalibrationSet(CalibrationSet, Adjustable):
         """
         return self._id
 
-    def save_output(self, i) -> None:
-        """
-            Save the last output to output for iteration i
-        """
-        #FIXME ensure _output_file exists
-        #FIXME re-enable this once more complete
-        shutil.move(self._output_file, '{}_last'.format(self._output_file))
-    
     #update handled in meta, TODO remove this method???
     def update_params(self, iteration: int) -> None:
         pass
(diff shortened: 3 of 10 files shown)


aaraney:
@shorvath-noaa if its not too much trouble, can you point to the commit or PR that removed support for `flowveldepth` in hdf5?

aaraney:
Can you check my understanding, @hellkite500? At some point, `t-route` stopped outputting `flowveldepth` in its hdf5 output, however that variable is still present in the csv format. _Both_ the csv and hdf5 outputs will be present in all `t-route` simulations (that output things) so this is not a breaking change, we are more or less just using a different data source? Just wanting to make sure that this is not breaking backwards compatibility.

shorvath-noaa:
> @shorvath-noaa if its not too much trouble, can you point to the commit or PR that removed support for `flowveldepth` in hdf5?

That was before my time, but I can search for it if that's helpful?

The latest output method for t-route is netcdf files containing all of the values from the flowveldepth array, plus info about streamflow nudging (DA values used). Would those work for this?

hellkite500:
> Can you check my understanding, @hellkite500? At some point, `t-route` stopped outputting `flowveldepth` in its hdf5 output, however that variable is still present in the csv format. _Both_ the csv and hdf5 outputs will be present in all `t-route` simulations (that output things) so this is not a breaking change, we are more or less just using a different data source? Just wanting to make sure that this is not breaking backwards compatibility.

Previously, the code looked for `flowveldepth.h5` and read the model output for evaluation from that file.  t-route no longer produces the output in hdf5 form, but it can produce the same table in csv form, so this PR just changes the source of the data it was looking for.  The selection of output format is a t-route configuration option and not directly the responsibility of ngen-cal, so I don't think this is a breaking change, per-se.

hellkite500:
@aaraney @shorvath-noaa I found some missing context...appearently t-route will still output hdf5 files -- if you use `main_v02` which in turn uses the v02 output function where this functionality was put in place to replace csv outputs with the binary hdf5:

https://github.com/NOAA-OWP/t-route/blob/1a5ccf243db939c186c0cdfe00c080b25cce5ab0/src/troute-nwm/src/nwm_routing/__main__.py#L845

However, this functionality was never refactored into v04 and its subsequent output mechanics, so it may be that we can revive this functionality with the addition of a new output configuration type/functionality.

I'm tempted to refactor this PR to look for hdf5 and revert to CSV and give users a choice based on their system/requirements.  Using hdf5 does require `tables` which has caused some minor hiccups in the past, but I think has been pretty stable and easy enough to use on most platforms/systems.

Thoughts on supporting either binary or text routing outputs in ngen-cal?

aaraney:
> I'm tempted to refactor this PR to look for hdf5 and revert to CSV and give users a choice based on their system/requirements. Using hdf5 does require `tables` which has caused some minor hiccups in the past, but I think has been pretty stable and easy enough to use on most platforms/systems.
> 
> Thoughts on supporting either binary or text routing outputs in ngen-cal?

If its not a crazy lift, this is my vote. Sounds like a little strategy pattern should get the job done!

hellkite500:
Note that this PR WILL break the explicit calibration semantics, a subsequent PR is in progress to refactor that strategy and bring it in alignment with these changes.

hellkite500:
Also probably need an updated example config illustrating the use of the model hooks, but that could be pushed in a subsequent PR...

python/ngen_cal/src/ngen/cal/_hookspec.py
```diff
@@ -42,3 +44,21 @@ def ngen_cal_finish(exception: Exception | None) -> None:
...
+    @hookspec(firstresult=True)
+    def ngen_cal_model_output(id: str | None) -> Series:
+        """
+            Called during each calibration iteration to provide the model output in the form
+            of a pandas Series, indexed by time.
+            Output series should be in units of cubic meters per second.
+        """
+
+    @hookspec
+    def ngen_cal_model_post_iteration(path: Path, iteration: int) -> None:
```
aaraney, line 58: What about these names instead?
```python
    def ngen_cal_iteration_trailing(path: Path, iteration: int) -> None: ...
    def ngen_cal_iteration_finish(path: Path, iteration: int) -> None: ...
    def ngen_cal_iteration_end(path: Path, iteration: int) -> None:
```

aaraney: I don't know why, but the `path` argument feels a little weird. It feels like its probably helpful in our current use cases, but I have a feeling there are other fields that could be useful. We might consider creating a [`typing.TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict) that gets passed in here and has a `path` field plus other things that we can add in the future.

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -108,6 +109,9 @@ def __init__(self, **kwargs):
         #Let pydantic work its magic
         super().__init__(**kwargs)
         #now we work ours
+        # Register the default ngen output hook
+        self._plugin_manager.register(TrouteOutput(self.routing_output))
```
aaraney, line 113: Maybe im just weird, but I think of the plugins like http middleware or handlers. Even if its a little "gross" I would rather have a single place in the codebase where all the plugins are registered if that is possible.

python/ngen_cal/src/ngen/cal/model.py
```diff
@@ -213,6 +217,23 @@ class ModelExec(BaseModel, Configurable):
...
+            type: type_as_import_string,
+            ModuleType: lambda mod: mod.__name__,
+            FunctionType: type_as_import_string,
+        }
+        underscore_attrs_are_private = True
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(**kwargs)
+        model_plugins = cast(List[Union[Callable, ModuleType]], self.plugins)
+        self._plugin_manager = setup_plugin_manager(model_plugins)
```
aaraney, line 236: I get the desire to setup the plugin manager here, but I would rather not. Why not just keep all the setup code in a few small easy to follow functions near `main`?

aaraney: Unrelated to my first comment, I am getting the feeling that we should just create a top level `plugins` and `plugin_settings` keys. So remove it from `general` and `model` and just move it to the top level. I think it will be the clearest and simplest to just have them all at the top level. Thoughts?

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
```diff
@@ -0,0 +1,83 @@
...
+            df.index = df.index.map(lambda x: 'wb-'+str(x))
+            df.columns = pd.MultiIndex.from_tuples(df.columns)
+        else:
+            df = None
+        return df
+    
+    # Try external provided output hooks, if those fail, try this one
+    # this will only execute if all other hooks return None (or they don't exist)
+    @hookimpl(specname="ngen_cal_model_output", trylast=True)    
+    def getOutput(self, id: str) -> Series:
```
aaraney, line 35: nitpick
```suggestion
    def get_output(self, id: str) -> Series:
```

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
```diff
@@ -0,0 +1,83 @@
+from ngen.cal import hookimpl
+
+import pandas as pd
+from pandas import DataFrame, Series
+from pathlib import Path
+
+class TrouteOutput():
```
aaraney, line 7: nitpick. I need to just get pre-commit hooks setup here so I don't have to be that guy lol.

```suggestion
class TrouteOutput:
```

python/ngen_cal/src/ngen/cal/_hookspec.py
```diff
@@ -42,3 +44,21 @@ def ngen_cal_finish(exception: Exception | None) -> None:
     raised during the calibration loop.
     `exception` will be non-none if an exception was raised during calibration.
     """
+
+class ModelHooks():
+    @hookspec(firstresult=True)
+    def ngen_cal_model_output(id: str | None) -> Series:
```
aaraney, line 50: Wondering if the weird behavior you were seeing with hooks returning more than a single value was because you are missing `self` here?

```suggestion
    def ngen_cal_model_output(self, id: str | None) -> Series:
```

python/ngen_cal/src/ngen/cal/_hookspec.py
aaraney, line 58: Same here:
```suggestion
    def ngen_cal_model_post_iteration(self, path: Path, iteration: int) -> None:
```

python/ngen_cal/src/ngen/cal/calibration_set.py
```diff
@@ -19,14 +21,16 @@ class CalibrationSet(Evaluatable):
         A HY_Features based catchment with additional calibration information/functionality
     """
 
-    def __init__(self, adjustables: Sequence[Adjustable], eval_nexus: Nexus, routing_output: 'Path', start_time: str, end_time: str, eval_params: 'EvaluationOptions'):
+    def __init__(self, adjustables: Sequence[Adjustable], eval_nexus: Nexus, hooks: 'HookRelay', start_time: str, end_time: str, eval_params: 'EvaluationOptions'):
```
aaraney, line 24: Why not just take in a `ModelHooks` instead? Then you get type hints and I think it is just generally more clear what is going on.

Aside, I think the coupling between the `CalibrationX` and `Evaluatable` are starting to show here. Not sure that we we can / should address that here, but I really want to push to try and flatten out the concepts in future additions. I think we will thank ourselves for that in the long run.

python/ngen_cal/src/ngen/cal/calibration_set.py
```diff
@@ -53,41 +57,22 @@ def output(self) -> 'DataFrame':
...
-        except Exception as e:
-            raise(e)
-        #if hydrograph is None:
-        #    raise(RuntimeError("Error reading output: {}".format(self._output_file)))
-        return hydrograph
-
+        # TODO should contributing_catchments be singular??? assuming it is for now...
+        # Call output hooks, take first non-none result provided from hooks (called in LIFO order of registration)
+        df = self._output_hook(id=self._eval_nexus.contributing_catchments[0].replace('cat', 'wb'))
+        if not df:
```
aaraney, line 63: We should be able to get rid of this once `self` is added to the `hookspec`s above.

python/ngen_cal/src/ngen/cal/calibration_set.py
```diff
@@ -134,11 +113,11 @@ class UniformCalibrationSet(CalibrationSet, Adjustable):
         A HY_Features based catchment with additional calibration information/functionality
     """
 
-    def __init__(self, eval_nexus: Nexus, routing_output: 'Path', start_time: str, end_time: str, eval_params: 'EvaluationOptions', params: dict = {}):
+    def __init__(self, eval_nexus: Nexus, hooks: 'HookRelay', start_time: str, end_time: str, eval_params: 'EvaluationOptions', params: dict = {}):
```
aaraney, line 116: Same here, why not just take a `ModelHooks`?

python/ngen_cal/src/ngen/cal/_hookspec.py
aaraney: Instead we could just pass a `JobMeta` here that has the `path`. I am a fan of that. I wouldnt be opposed to moving `JobMeta` to a `typing.TypedDict` either.

aaraney (commented):
In general, i'm really happy with how using the plugin system decouples the system and really decomposes things. I think that is a huge win and as we continue down the road, I know we will thank ourselves for going this sort of route.

I left a few comments that should be pretty straight forward to get though. It might be easiest if we found some time tomorrow to sit down and work through them together. Thanks for spending time to think through this stuff!

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
```diff
@@ -0,0 +1,83 @@
+from ngen.cal import hookimpl
```
aaraney, line 1: This should fix this test failures!

```suggestion
from __future__ import annotations
```

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: Looking at the types provided by the plugin manager, I could extract a single hook as a `HookCaller` or all hooks managed by the manager as a `HookRelay` and extract the callers as needed, I opted for the latter to allow more hooks to be added/used as needed.

python/ngen_cal/src/ngen/cal/_hookspec.py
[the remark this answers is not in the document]
hellkite500: Perhaps a slightly different name?  I think we still need something specific to extract model outputs into a time series which will be evaluated.  We may in the future also need to provide some time bounds to this at some point as well (in the future, I think...)

python/ngen_cal/src/ngen/cal/ngen.py
hellkite500: This has several nuanced considerations, with valid concerns on both sides of the topic.  For now, I propose leaving this in and seeing how more hook semantics play out and how we might want to manage globally vs scoped/controlled instances.

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: I'll test this out and see what happens.

python/ngen_cal/src/ngen/cal/_hookspec.py
aaraney: What about `ngen_cal_iteration_finish(iteration: int, job_meta: JobMeta)`?

python/ngen_cal/src/ngen/cal/_hookspec.py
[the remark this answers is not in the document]
aaraney: What about just: 

```python
    def ngen_cal_model_output(self) -> pd.DataFrame: ...
```

python/ngen_cal/src/ngen/cal/_hookspec.py
hellkite500: renamed and adjusted to use job meta in 0b272d088de21ca23e6c35ff804a7aa004e3da56

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
hellkite500: fixed in 082e2712272f339a94c5d01fa8a99f33a553b1ea

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: Still get back a list object
```
[2018-01-01 00:00:00        0.538083
2018-01-01 01:00:00      789.833000
2018-01-01 02:00:00     8354.108000
2018-01-01 03:00:00    13602.951000
2018-01-01 04:00:00    13904.544000
2018-01-01 05:00:00    13648.108000
2018-01-01 06:00:00    13262.755000
2018-01-01 07:00:00    12760.119000
2018-01-01 08:00:00    12311.113000
2018-01-01 09:00:00    11847.882000
2018-01-01 10:00:00    11454.423000
2018-01-01 11:00:00    11056.809000
2018-01-01 12:00:00    10716.239000
2018-01-01 13:00:00    10372.783000
2018-01-01 14:00:00    10075.482000
2018-01-01 15:00:00     9775.703000
2018-01-01 16:00:00     9513.855000
2018-01-01 17:00:00     9249.671000
2018-01-01 18:00:00     9017.308000
2018-01-01 19:00:00     8797.409000
2018-01-01 20:00:00     8575.065000
2018-01-01 21:00:00     8377.901000
2018-01-01 22:00:00     8178.397000
2018-01-01 23:00:00     8000.547400
Freq: h, Name: sim_flow, dtype: float64]
<class 'list'>
```

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: Meta Duck Typing 🫨 will try it out I suppose 😄

python/ngen_cal/src/ngen/cal/_hookspec.py
hellkite500: Fixed in 082e2712272f339a94c5d01fa8a99f33a553b1ea, behavior didn't change...

python/ngen_cal/src/ngen/cal/_hookspec.py
hellkite500: Fixed in 082e2712272f339a94c5d01fa8a99f33a553b1ea

python/ngen_cal/src/ngen/cal/_hookspec.py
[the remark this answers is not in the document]
hellkite500: The id is required context in some form to allow the function to only provide output for the feature being evaluated.  We could load the entire output into a dataframe and subset it on the other side, but that is potentially a lot of wasted I/O and memory.  Also, the semantics of which output to load and return would need to be communicated to the hook as well, whereas restricting to a single time series of information for a feature leaves the choice up to the implementer which variable (stream flow, catchment variable, ect...) to return.

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
hellkite500: changed in c33327b72e1c60a42e03cc1d2f7d22f6e745dafe

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
hellkite500: fixed in c33327b72e1c60a42e03cc1d2f7d22f6e745dafe

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: changed in 2602282227192d6b4703fde0cf155c1a2c003a75

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: changed in 2602282227192d6b4703fde0cf155c1a2c003a75

python/ngen_cal/src/ngen/cal/calibration_set.py
```diff
@@ -19,14 +25,16 @@ class CalibrationSet(Evaluatable):
...
+    def __init__(self, adjustables: Sequence[Adjustable], eval_nexus: Nexus, hooks: ModelHooks, start_time: str, end_time: str, eval_params: 'EvaluationOptions'):
         """
 
         """
         super().__init__(eval_params)
         self._eval_nexus = eval_nexus
         self._adjustables = adjustables
-        self._output_file = routing_output
+        # record the hooks needed for output and checkpointing
+        self._output_hook = hooks.ngen_cal_model_output
```
aaraney, line 36: Just to clean this up a bit, why not just set:

```python
        self._hooks: ModelHooks = hooks
```
and use that where it is needed. Just a little easier to follow. I for sure would be wondering why there was an added layer of indirection if I were to stumble on this code.

python/ngen_cal/src/ngen/cal/calibration_set.py
aaraney: That's super weird. I really don't think that should be happening and is worth us figuring out why it is returning a list instead of a `pd.Series` as expected.

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: fair point, changed in 308330ad0a01be638b7757d6eae3f064fbf2f441

python/ngen_cal/src/ngen/cal/_hookspec.py
```diff
@@ -42,3 +45,21 @@ def ngen_cal_finish(exception: Exception | None) -> None:
     raised during the calibration loop.
     `exception` will be non-none if an exception was raised during calibration.
     """
+
+class ModelHooks():
```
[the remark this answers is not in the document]
hellkite500: c8962e536d958e2d2d605becd481417db2d15774

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
```diff
@@ -0,0 +1,95 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ngen.cal.meta import JobMeta
+
+class TrouteOutput:
+
+    def __init__(self, filepath: Path) -> None:
+        self._output_file = filepath
```
aaraney, line 21: Thoughts on doing something like:

```python
assert filepath.exist(), f"filepath: {str(filepath)} does not exits"
assert filepath.suffix.lower() in (".csv", ".hdf5"), f"unsupported file type: {filepath.suffix}"
```

I think then we could clean up the return type of `_get_dataframe` to just be a `DataFrame`. To me, that just makes the invariants really clear.

hellkite500: We use the `None` type to indicate to the application that an output doesn't exist so it should attempt to run the model to produce an initial output...

aaraney: Fair enough. I didn't think this would run before a model ever would.

python/ngen_cal/src/ngen/cal/_hookspec.py
```diff
@@ -42,3 +45,21 @@ def ngen_cal_finish(exception: Exception | None) -> None:
...
+    @hookspec(firstresult=True)
+    def ngen_cal_model_output(self, id: str | None) -> Series:
+        """
+            Called during each calibration iteration to provide the model output in the form
+            of a pandas Series, indexed by time.
+            Output series should be in units of cubic meters per second.
+        """
+
+    @hookspec
+    def ngen_cal_model_iteration_finish(self, info: JobMeta, iteration: int) -> None:
```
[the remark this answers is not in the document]
hellkite500: changed in 93afea14cf1788b4cc5213876b3baf989beab230

python/ngen_cal/src/ngen/cal/calibratable.py
```diff
@@ -80,10 +83,11 @@ def check_point_file(self) -> 'Path':
         """
         return Path('{}_parameter_df_state.parquet'.format(self.id))
 
-    def check_point(self, path: 'Path') -> None:
+    def check_point(self, info: JobMeta, iteration: int) -> None:
```
aaraney, line 86: Might wanna switch this around too

python/ngen_cal/src/ngen/cal/calibration_set.py
```diff
@@ -106,20 +94,14 @@ def observed(self) -> 'DataFrame':
...
-    def save_output(self, i) -> None:
-        """
-            Save the last output to output for iteration i
-        """
-        #FIXME ensure _output_file exists
-        #FIXME re-enable this once more complete
-        shutil.move(self._output_file, '{}_last'.format(self._output_file))
-    
-    def check_point(self, path: 'Path') -> None:
+    def check_point(self, info: JobMeta, iteration: int) -> None:
```
aaraney, line 97: Same here

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -127,7 +125,7 @@ def dds(start_iteration: int, iterations: int,  calibration_object: 'Evaluatable
             _execute(agent)
         with pushd(agent.job.workdir):
             _evaluate(0, calibration_object, info=True)
-        calibration_object.check_point(agent.job.workdir)
+        calibration_object.check_point(agent.job, 0)
```
aaraney, line 128:
```suggestion
        calibration_object.check_point(0, agent.job)
```

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -139,7 +137,7 @@ def dds(start_iteration: int, iterations: int,  calibration_object: 'Evaluatable
         _execute(agent)
         with pushd(agent.job.workdir):
             _evaluate(i, calibration_object, info=True)
-        calibration_object.check_point(agent.job.workdir)
+        calibration_object.check_point(agent.job, i)
```
aaraney, line 140:
```suggestion
        calibration_object.check_point(i, agent.job)
```

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -177,7 +175,7 @@ def dds_set(start_iteration: int, iterations: int, agent: 'Agent'):
                 _execute(agent)
             with pushd(agent.job.workdir):
                 _evaluate(0, calibration_set, info=True)
-            calibration_set.check_point(agent.job.workdir)
+            calibration_set.check_point(agent.job, 0)
```
aaraney, line 178:
```suggestion
            calibration_set.check_point(0, agent.job)
```

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -190,7 +188,7 @@ def dds_set(start_iteration: int, iterations: int, agent: 'Agent'):
             _execute(agent)
             with pushd(agent.job.workdir):
                 _evaluate(i, calibration_set, info=True)
-            calibration_set.check_point(agent.job.workdir)
+            calibration_set.check_point(agent.job, i)
```
aaraney, line 191:
```suggestion
            calibration_set.check_point(i, agent.job)
```

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -204,7 +202,7 @@ def compute(calibration_object, iteration, input) -> float:
         agent.update_config(iteration, calibration_object.df[[str(iteration), 'param', 'model']], calibration_object.id)
         _execute(agent)
         cost = _evaluate(iteration, calibration_object)
-        calibration_object.check_point(agent.workdir)
+        calibration_object.check_point(agent.job, iteration)
```
aaraney, line 205:
```suggestion
        calibration_object.check_point(iteration, agent.job)
```

python/ngen_cal/src/ngen/cal/search.py
```diff
@@ -274,7 +272,7 @@ def pso_search(start_iteration: int, iterations: int,  agent):
         #For pyswarm, DO NOT use the embedded multi-processing -- it is impossible to track the mapping of an agent to the params
         cost, pos = optimizer.optimize(cf, iters=iterations, n_processes=None)
         calibration_object.df.loc[:,'global_best'] = pos
-        calibration_object.check_point("./")
+        calibration_object.check_point(agent.job, iterations)
```
aaraney, line 275:
```suggestion
        calibration_object.check_point(iterations, agent.job)
```

python/ngen_cal/tests/test_model.py
```diff
@@ -58,9 +59,9 @@ def test_restart_1(ngen_config: 'Ngen', eval: 'EvaluationOptions', workdir: 'Dir
     eval._best_params_iteration = "1"
     ngen_config.adjustables[0]._best_params = "1"
     eval.write_param_log_file(2)
-    
+    info = JobMeta(ngen_config.type, workdir, workdir = workdir)
     #make sure the catchment param df is saved before trying to restart
-    ngen_config.adjustables[0].check_point(workdir)
+    ngen_config.adjustables[0].check_point(info, 1)
```
aaraney, line 64:
```suggestion
    ngen_config.adjustables[0].check_point(1, info)
```

python/ngen_cal/tests/test_plugin_system.py
```diff
@@ -49,6 +50,13 @@ def ngen_cal_start() -> None:
 def ngen_cal_finish() -> None:
     """Called after exiting the calibration loop."""
 
+@hookimpl
+def ngen_cal_model_output() -> None:
+    """Test model output plugin"""
+
+@hookimpl
+def ngen_cal_model_post_iteration(path: Path, iteration: int) -> None:
```
aaraney, line 58:
```suggestion
def ngen_cal_model_post_iteration(iteration: int, info: JobMeta) -> None:
```

python/ngen_cal/tests/test_plugin_system.py
```diff
@@ -61,3 +69,11 @@ def ngen_cal_start(self) -> None:
     @hookimpl
     def ngen_cal_finish(self) -> None:
         """Called after exiting the calibration loop."""
+
+    @hookimpl
+    def ngen_cal_model_output(self) -> None:
+        """Test model output plugin"""
+
+    @hookimpl
+    def ngen_cal_model_post_iteration(self, path: Path, iteration: int) -> None:
```
aaraney, line 78:
```suggestion
    def ngen_cal_model_post_iteration(self, iteration: int, info: JobMeta) -> None:
```

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: Create an issue and leave the workaround for now?

python/ngen_cal/src/ngen/cal/calibratable.py
hellkite500: now you really are getting picky 🤣

hellkite500: changed in 50b3f05d93dd96d8761a421e060f02f07b86bc00

python/ngen_cal/src/ngen/cal/calibration_set.py
hellkite500: 50b3f05d93dd96d8761a421e060f02f07b86bc00

python/ngen_cal/tests/test_plugin_system.py
hellkite500: History has gotten out of sync, this function was changed to `...iteration_finish` several commits back 😄

python/ngen_cal/src/ngen/cal/calibration_set.py
aaraney: Sounds good and like something I can look into once we get this in. Opened https://github.com/NOAA-OWP/ngen-cal/issues/135

python/ngen_cal/tests/test_model.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/search.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/search.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/ngen_hooks/ngen_output.py
aaraney: > "We" being me

😆

python/ngen_cal/src/ngen/cal/search.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/search.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/search.py
hellkite500: 658b5535b096d57ef92b43b40ace296467fc3837

python/ngen_cal/src/ngen/cal/_hookspec.py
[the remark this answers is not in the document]
hellkite500: Will come back to this in future work as model/session info structures are more flushed out.

aaraney (approved):
Looks good to me! Thanks for working through this with me, @hellkite500! Lets squash and merge this puppy!

python/ngen_cal/src/ngen/cal/model.py
hellkite500: will leave this for now, and keep in mind plugin management, scope, ect as we continue to build on the concept.

Issue #135, opened 2024-07-24
Investigate why `ngen_cal_model_output` returns a `list[pd.Series]` instead of a `pd.Series`


aaraney:
_Originally posted by @aaraney in https://github.com/NOAA-OWP/ngen-cal/pull/130#discussion_r1690032175_

aaraney:
This ended up being a bug in either the way the `hookspec` was defined or it was being called. This has been resolved.

PR #137, opened 2024-07-24
Pso fix

hellkite500:
Quick fix to workaround multiprocessing/pickle interactions with the plugin manager for PSO optimization.  Also relevant to #135.

### DIFF
--- a/python/ngen_cal/src/ngen/cal/_plugin_system.py
+++ b/python/ngen_cal/src/ngen/cal/_plugin_system.py
@@ -28,15 +28,14 @@ def setup_plugin_manager(plugins: list[Callable | ModuleType]) -> PluginManager:
 
     return pm
 
-def setup_scoped_plugin_manager(spec: type, plugins: list[Callable | ModuleType]) -> PluginManager:
+def setup_scoped_plugin_manager(spec: type, plugins: list[Callable]) -> PluginManager:
     pm = pluggy.PluginManager(PROJECT_SLUG)
     pm.add_hookspecs(spec)
 
     for plugin in plugins:
+        assert not isinstance(plugin, ModuleType), "function plugins"
         if isinstance(plugin, Callable):
             pm.register(plugin())
-        elif isinstance(plugin, ModuleType):
-            pm.register(plugin)
         else:
             assert_never(plugin)
 
--- a/python/ngen_cal/src/ngen/cal/_plugin_system.py
+++ b/python/ngen_cal/src/ngen/cal/_plugin_system.py
@@ -27,3 +27,17 @@ def setup_plugin_manager(plugins: list[Callable | ModuleType]) -> PluginManager:
             assert_never(plugin)
 
     return pm
+
+def setup_scoped_plugin_manager(spec: type, plugins: list[Callable | ModuleType]) -> PluginManager:
+    pm = pluggy.PluginManager(PROJECT_SLUG)
+    pm.add_hookspecs(spec)
+
+    for plugin in plugins:
+        if isinstance(plugin, Callable):
+            pm.register(plugin())
+        elif isinstance(plugin, ModuleType):
+            pm.register(plugin)
+        else:
+            assert_never(plugin)
+
+    return pm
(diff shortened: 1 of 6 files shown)


python/ngen_cal/src/ngen/cal/_plugin_system.py
```diff
@@ -27,3 +27,17 @@ def setup_plugin_manager(plugins: list[Callable | ModuleType]) -> PluginManager:
             assert_never(plugin)
 
     return pm
+
+def setup_scoped_plugin_manager(spec: type, plugins: list[Callable | ModuleType]) -> PluginManager:
```
aaraney, line 31:
```suggestion
def setup_namespaced_plugin_manager(spec: type, plugins: list[Callable]) -> PluginManager:
```

I think the type hints here can be cleaned up slightly just emphasize the invariants at play.

python/ngen_cal/src/ngen/cal/_plugin_system.py
```diff
@@ -27,3 +27,17 @@ def setup_plugin_manager(plugins: list[Callable | ModuleType]) -> PluginManager:
...
+    pm = pluggy.PluginManager(PROJECT_SLUG)
+    pm.add_hookspecs(spec)
+
+    for plugin in plugins:
+        if isinstance(plugin, Callable):
+            pm.register(plugin())
+        elif isinstance(plugin, ModuleType):
```
aaraney, line 38: If a module based plugin is registered, we will run into the same pickling a module issue in a multiprocessing context. I think we should just `assert not isinstance(plugin, ModuleType), "function plugins"` and catch that at the call site and re-raise with extra context.

python/ngen_cal/src/ngen/cal/model.py
```diff
@@ -233,7 +234,7 @@ class Config(BaseModel.Config):
     def __init__(self, *args, **kwargs):
         super().__init__(**kwargs)
         model_plugins = cast(List[Union[Callable, ModuleType]], self.plugins)
-        self._plugin_manager = setup_plugin_manager(model_plugins)
+        self._plugin_manager = setup_scoped_plugin_manager(ModelHooks, model_plugins)
```
aaraney, line 237:
```suggestion
        try:
            self._plugin_manager = setup_scoped_plugin_manager(ModelHooks, model_plugins)
        except AssertionError as e:
            raise RuntimeError("At this time, model plugins do not support function based plugins.")
```

python/ngen_cal/src/ngen/cal/_plugin_system.py
hellkite500: changed in a4710a941b0869d1bf1c2d2bfd920b93b09646fc

python/ngen_cal/src/ngen/cal/_plugin_system.py
hellkite500: refactored in a4710a941b0869d1bf1c2d2bfd920b93b09646fc

aaraney (approved):
Thanks, @hellkite500!

aaraney (changes requested):
Looks like one of the tests if correctly failing now.

https://github.com/NOAA-OWP/ngen-cal/actions/runs/10096240522/job/27918258512?pr=137#step:7:91

aaraney:
Looks like these test function plugins need to be removed now (https://github.com/NOAA-OWP/ngen-cal/blob/61c9672559832d5bb2edf7471f8d1c1cf959f9c6/python/ngen_cal/tests/test_plugin_system.py#L53-L59).

hellkite500:
The problem was that I modifed `setup_plugin_manager` instead of `setup_scoped_plugin_manager` haha, I just corrected that.

aaraney:
> The problem was that I modifed `setup_plugin_manager` instead of `setup_scoped_plugin_manager` haha, I just corrected that.

Nice!

aaraney:
In doing some local testing, we should be able to get around some of this if we use [`multiprocess`](https://github.com/uqfoundation/multiprocess?tab=readme-ov-file#multiprocess). It uses [`dill`](https://github.com/uqfoundation/dill) an alternative to pickle, to serialize and deserialize objects.

PR #138, opened 2024-07-30
fix / feat: handle building gpkg crosswalk from `flowpath-attributes`

aaraney:
Newer HF versions now use:
`flowlines` instead of `flowpaths`
`flowpath-attributes` instead of `flowpath_attributes`
`model-attributes` instead of `model_attributes`
This just adds support for that.

## Additions

- Add support for hydrofabrics that use `flowlines` and `flowpath-attributes`

### DIFF
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -117,16 +117,34 @@ class NgenBase(ModelExec):
        
         # Read the catchment hydrofabric data
         if self.hydrofabric is not None:
-            self._read_hydrofabric()
+            if self._is_legacy_gpkg_hydrofabric(self.hydrofabric):
+                self._read_legacy_gpkg_hydrofabric()
+            else:
+                self._read_gpkg_hydrofabric()
         else:
-            self._read_legacy_hydrofabric()
+            self._read_legacy_geojson_hydrofabric()
 
         #Read the calibration specific info
         with open(self.realization) as fp:
             data = json.load(fp)
         self.ngen_realization = NgenRealization(**data)
 
-    def _read_hydrofabric(self):
+    @staticmethod
+    def _is_legacy_gpkg_hydrofabric(hydrofabric: Path) -> bool:
+        """Return True if legacy hydrofabric."""
+        import sqlite3
+        connection = sqlite3.connect(hydrofabric)
+        # hydrofabric <= 2.1 use 'flowpaths'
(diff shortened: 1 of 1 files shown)


hellkite500:
Which version uses which convention?

aaraney:
2.1.1 - https://www.lynker-spatial.com/data?path=hydrofabric%2Fv2.1.1%2Fnextgen%2F

aaraney:
Re-running jobs on master b.c. I don't think they are failing from what i've added here.

aaraney:
Opened #139 to track the test failures. Confirmed the additions made here.

aaraney:
Having chatted with @mikejohnson51 (correct me where i'm wrong), all HF releases `>2.1` `flowpaths` -> `flowlines` and `flowpath_attributes` -> `flowpath-attributes`. Can you confirm that @mikejohnson51?

python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -115,45 +115,94 @@ def __init__(self, **kwargs):
...
+            self._read_legacy_geojson_hydrofabric()
 
         #Read the calibration specific info
         with open(self.realization) as fp:
             data = json.load(fp)
         self.ngen_realization = NgenRealization(**data)
 
+    @staticmethod
+    def _is_legacy_gpkg_hydrofabric(hydrofabric: Path) -> bool:
+        """Return True if legacy hydrofabric."""
```
robertbartel, line 134: It would be helpful to explain somewhere (i.e., either here, or with a reference here to wherever that is) what constitutes a "legacy hydrofabric."  Are we talking about versions prior to v2.0?  Versions >= 2.0 but older than the latest?  I could make an educated guess, but, even assuming a high probability of correctness right now, that will get harder for someone in the future to do reliably.

robertbartel (approved):
I had one comment on perhaps expanding some on the docstring to clarify what constitutes a "legacy hydrofabric," but this isn't critical and may be tricky given how hydrofabric data and schemas are currently versioned.  So I'm good with it as-is.

python/ngen_cal/src/ngen/cal/ngen.py
aaraney: Nice suggestion. Just updated the docstring to better reflect what it means to be legacy.

aaraney:
@robertbartel, just addressed your comment. Should be good to go now. Thanks!

robertbartel (approved):

PR #142, opened 2024-07-31
fix: allow netcdf forcing

### LINKED ISSUE #134: Allow ngen realizations with NetCDF data providers for forcing
Calibration runs should work with realizations using both CSV and NetCDF providers.

## Current behavior

When using NetCDF provider with an Independent Strategy, an error occurs when ngen_cal tries to create the indepdent catchment definitions in the new realization.  Currently, it tries to create a custom forcing definition and assumes a file pattern substituion is required.  But ngen doesn't allow this kind of structure with NetCDF Providers, so the keys doesn't exist correctly in the global formulation definition which causes an error.

## Expected behavior

Create proper independent realization configurations from input realizations that have a NetCDF provider defined for the global formulation.

aaraney:
closes #134

Add support for forcing files without patterns (e.g. `netcdf`).

### DIFF
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -437,14 +437,16 @@ class NgenIndependent(NgenBase):
             #Need to fix the forcing definition or ngen will not work
             #for individual catchment configs, it doesn't apply pattern resolution
             #and will read the directory `path` key as the file key and will segfault
-            pattern = catchment_realizations[id].forcing.file_pattern
             path = catchment_realizations[id].forcing.path
+            pattern = catchment_realizations[id].forcing.file_pattern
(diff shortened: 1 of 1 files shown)


aaraney:
TODO: drop unrelated commits and rebase.

aaraney:
I had @ajkhattak run a test to verify that it worked. Still need to test it out for myself. 

Aside, I would really like to prioritize getting integration tests setup in CI so this could just become a test case.

ajkhattak:
I think I can run a quick test to check if it works with both `.nc` and `.csv` formats, if that would help.

aaraney:
@ajkhattak, yeah that would be great. Thanks for offering!

aaraney:
@ajkhattak, tested and was able to run the model

hellkite500 (approved):

PR #156, opened 2024-08-05
chore: remove `calibration.py`

### LINKED ISSUE #54: Deprecate calibration.py
The [calibration.py](https://github.com/NOAA-OWP/ngen-cal/blob/master/python/calibration.py) file is a second interface to the `ngen-cal` package, and the functionality has been migrated to the package `__main__.py` module.  This should probably get deprecated and removed as the calibration entry point and users/workflows should update to calling the package directly.

aaraney:
closes #54


hellkite500:
Can we do a quick check on any documentation, examples, ect that may reference this and update them to point to the package main?

aaraney:
> Can we do a quick check on any documentation, examples, ect that may reference this and update them to point to the package main?

Yep, I couldn't find any in the OWP or other relevant orgs. That is why I felt confident that we could remove it.

aaraney:
All the folks that, at least I know, using `ngen.cal` using the `__main__.py` entrypoint.

hellkite500 (approved):

PR #157, opened 2024-08-06
update style to more modern python idioms and remove dead imports

aaraney:
This PR does introduce behavioral changes. Its highly recommended to view each commit individually during review. Ive split things up by the refactoring / stylistic operation that was performed to make this easier to review.

The refactors / stylistic improvements performed are:

chore:
- remove unused imports
- add `from __future__ import annotations` to relevant files

refactor: 
- guard non-runtime types with `typing.TYPE_CHECKING`
- single member `typing.Union`

style: 
- use `__future__` `annotations` and remove quotes from, now, non-runtime type hints
- drop open's default 'r'
- use `f`-strings

Note: non-`typing` types (e.g. `list[str]`) do not work with `pydantic` for `python<=3.8`. [See](https://github.com/pydantic/pydantic/issues/2453#issuecomment-789110628) this issue for more information.


python/ngen_config_gen/src/ngen/config_gen/models/cfe.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from typing import Any, Dict, List, Union
```
hellkite500: Since things like `Dict` and `List` were changed to `dict` and `list` are this imports required still?  This would be relevant across (potentially) all the files changed in this commit...

python/ngen_cal/tests/test_configuration.py
```diff
@@ -1,30 +0,0 @@
...
-# def test_config_file(conf: 'Configuration', realization_config: str) -> None:
-#     """
-#         Test configuration property `config_file`
-#     """
-#     assert conf.config_file == realization_config
-
-# @pytest.mark.usefixtures("conf")
-# def test_catchments(conf: 'Configuration') -> None:
-#     """
-#         Ensure that only the catchment marked with "calibration" is used in the configuration
```
hellkite500, line 27: I don't recall why these tests were all commented, but do you see any value in reviving any of them or some form of them?

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import TYPE_CHECKING, Any, Generic, List, Optional, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Generic, List, TypeVar
```
hellkite500, line 4: Earlier comment about imports is being at least partially resolved in this commit (review by commit as suggested, so didn't see this at the time.). However, I still see `List` where `list` was used in the previous commit, is `List` still needed?  Should probably double check consistent useage of `list` and `List`

python/ngen_conf/src/ngen/config/init_config/noahowp.py
```diff
@@ -4,7 +4,7 @@
 from datetime import datetime
 from enum import Enum
 from pathlib import Path, PosixPath, WindowsPath
-from typing import ClassVar, Dict, List, Literal, Union
+from typing import ClassVar, List, Literal
```
hellkite500, line 7: Another place to double check `List`

python/ngen_cal/tests/conftest.py
```diff
@@ -151,12 +151,12 @@ def catchment(nexus, fabric, workdir, mocker) -> Generator[CalibrationCatchment,
     output = nexus._hydro_location.get_data().rename(columns={'value':'sim_flow'})
     output.set_index('value_time', inplace=True)
     #Override the output property so it doesn't try to reload output each time
-    mocker.patch(__name__+'.EvaluatableCatchment.output',
+    mocker.patch('ngen.cal.calibration_cathment.EvaluatableCatchment.output',
```
hellkite500, line 154: Does the work to run tests without having an editable install?

python/ngen_conf/src/ngen/config/configurations.py
```diff
@@ -22,9 +22,9 @@ class Provider(str, Enum):
     file_pattern: Optional[Union[Path, str]]
     path: Path
     #reasonable? default
-    provider: Provider = Field(Provider.CSV)
+    provider: Forcing.Provider = Field(Provider.CSV)
```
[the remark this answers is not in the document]
aaraney: Pydantic... This is where this change set gets a little confusing. It comes down to the scoping of non-runtime type hints. So, in this case, `provider: Provider` causes an issue.

I think the most sane thing to do to avoid confusion and move `Provider` out of the class.

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
```diff
@@ -65,14 +65,14 @@ def _serialize(self) -> str:
     def dict(
         self,
         *,
-        include: Optional[Union["AbstractSetIntStr", "MappingIntStrAny"]] = None,
-        exclude: Optional[Union["AbstractSetIntStr", "MappingIntStrAny"]] = None,
+        include: AbstractSetIntStr | MappingIntStrAny | None = None,
```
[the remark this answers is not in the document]
aaraney: Right, sooooo much better

python/ngen_conf/src/ngen/config/path_pair/path_pair.py
```diff
@@ -137,9 +137,9 @@ def __new__(
                     f"Filename not derived from template and pattern, {template_str.name!r} {pattern!r}: {item}"
                 )
 
-        self: Union[
-            WindowsPathPairCollection[T], PosixPathPairCollection[T]
-        ] = Path.__new__(
+        self: (
```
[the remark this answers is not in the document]
aaraney: Yeah

python/ngen_config_gen/src/ngen/config_gen/models/cfe.py
aaraney: Yeah, I didn't change the `pydantic` model type hints and I really should.
We will still have to use `typing.Optional` and `typing.Union` in `pydantic` model's for now, but we should be able to remove the others (I think). 

More plainly, _pydantic_ can't use `field: Foo | None = None` or `field: Foo | Bar` with `python<3.10`. See https://github.com/pydantic/pydantic/issues/2453#issuecomment-789110628

python/ngen_cal/tests/test_configuration.py
aaraney: Probably, Ill open an issue to track re-adding them. For now, lets axe dead code though.

aaraney: See https://github.com/NOAA-OWP/ngen-cal/issues/165

python/ngen_conf/src/ngen/config/init_config/value_unit_pair.py
aaraney: Yeah this is certainly a little confusing. When I was making these changes, I was under the impression that `pydantic` model's still needed to be annotated with types from the `typing` package. I was wrong about this... Only `Union` and `Optional` need to stay as _runtime_ type hints in pydantic modules.

python/ngen_cal/tests/conftest.py
aaraney: Works in both cases.

hellkite500: Better question -- if a fresh clone is pulled, and you run the tests without doing any install, will they work?

hellkite500: Required when using a `src` layout, it seems.  We should revisit this general topic, but it doesn't have to limit this PR if the package imports in other parts of the test code require an install to work.

hellkite500: If possible, I suggest moving this to its own change set and/or issue.  It seems out of scope with the dead imports and style of the rest of this PR.

aaraney: I created #166 to track this. Putting it on my todos

aaraney:
@hellkite500, I think we should be good to merge this now.

hellkite500 (dismissed):

hellkite500 (approved):

PR #158, opened 2024-08-07
Add `pre-commit`, `pre-commit` gh action, and update contribution guide

aaraney:
Add [`pre-commit`](https://pre-commit.com/) configuration, a gh ci action that enforces `pre-commit` rules, and update contribution guide.

See: https://pre-commit.com/index.html

Highly recommend viewing individual commits when reviewing this. I used `pre-commit` to remove trailing white spaces and add missing trailing newlines.

In the future I plan to add linting and formatting stages using [`ruff`](https://github.com/astral-sh/ruff-pre-commit). This seemed a little overly disruptive for now.

PR #159, opened 2024-08-07
Add `pre-commit` ci action

aaraney:
DON'T MERGE: still testing


aaraney:
Feeling like i'm on the edge of perfection being the enemy of progress. In the future I would like a CI action that you can interact with via comments to auto-fix issues that the `pre-commit` action finds. This seems doable with something like [pre-commit lite](https://pre-commit.ci/lite.html), however it requires that we install a GitHub app that can interact with the repo. The app requires org approval and i've not done enough research to start going down that path (possible there are other alternatives). Closing this for now and introducing the `pre-commit` GH action in #158.

PR #161, opened 2024-08-09
feat: `ngen_cal_model_configure` model hook

aaraney:
Add model hook: `ngen_cal_model_configure` hook that mirrors the existing `ngen_cal_configure` hook.

```python
    @hookspec
    def ngen_cal_model_configure(self, config: ngen.cal.model.ModelExec) -> None:
        """
        Called before calibration begins.
        This allow plugins to perform initial configuration.

        Plugins' configuration data should be provided using the
        `plugins_settings` field in the `model` section of an `ngen.cal`
        configuration file.
        By convention, the name of the plugin should be used as top level key in
        the `plugin_settings` dictionary.
        """
```
## Additions

- `ngen_cal_model_configure` model hook; Called before calibration begins. This allow plugins to perform initial configuration.
- `unwrap` method added to `ngen.cal.ngen.Ngen`. Returns the instance's `__root__` attribute.

## Changes

- `Agent` now takes a `Model` instance as a parameter. Previously, the raw config was passed.

### DIFF
--- a/python/ngen_cal/src/ngen/cal/__main__.py
+++ b/python/ngen_cal/src/ngen/cal/__main__.py
@@ -4,7 +4,8 @@ from __future__ import annotations
 import yaml
 from os import chdir
 from pathlib import Path
-from ngen.cal.configuration import General
+from ngen.cal.configuration import General, Model
+from ngen.cal.ngen import Ngen
 from ngen.cal.search import dds, dds_set, pso_search
 from ngen.cal.strategy import Algorithm
 from ngen.cal.agent import Agent
@@ -42,6 +43,13 @@ def main(general: General, model_conf: Mapping[str, Any]):
         import numpy as np
         np.random.seed(general.random_seed)
 
+    # model scope plugins setup in constructor
+    model = Model(model=model_conf)
+
+    # NOTE: if support for new models is added, this will need to be modified
+    assert isinstance(model.model, Ngen), f"ngen.cal.ngen.Ngen expected, got {type(model.model)}"
+    model_inner = model.model.unwrap()
+
     plugins = cast(List[Union[Callable, ModuleType]], general.plugins)
     plugin_manager = setup_plugin_manager(plugins)
 
@@ -57,8 +65,9 @@ def main(general: General, model_conf: Mapping[str, Any]):
     into a single variable vector and calibrating a set of heterogenous formultions...
     """
     start_iteration = 0
-    #Initialize the starting agent
-    agent = Agent(model_conf, general.workdir, general.log, general.restart, general.strategy.parameters)
+
+    # Initialize the starting agent
+    agent = Agent(model, general.workdir, general.log, general.restart, general.strategy.parameters)
     if general.strategy.algorithm == Algorithm.dds:
         func = dds_set #FIXME what about explicit/dds
         start_iteration = general.start_iteration
--- a/python/ngen_cal/src/ngen/cal/__main__.py
+++ b/python/ngen_cal/src/ngen/cal/__main__.py
@@ -57,6 +57,7 @@ def main(general: General, model_conf: Mapping[str, Any]):
 
     # setup plugins
     plugin_manager.hook.ngen_cal_configure(config=general)
+    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
 
     print("Starting calib")
 
--- a/python/ngen_cal/src/ngen/cal/__main__.py
+++ b/python/ngen_cal/src/ngen/cal/__main__.py
@@ -57,7 +57,6 @@ def main(general: General, model_conf: Mapping[str, Any]):
 
     # setup plugins
     plugin_manager.hook.ngen_cal_configure(config=general)
-    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
 
     print("Starting calib")
 
@@ -69,6 +68,10 @@ def main(general: General, model_conf: Mapping[str, Any]):
 
     # Initialize the starting agent
     agent = Agent(model, general.workdir, general.log, general.restart, general.strategy.parameters)
+
+    # Agent mutates the model config, so `ngen_cal_model_configure` is called afterwards
+    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
+
     if general.strategy.algorithm == Algorithm.dds:
         func = dds_set #FIXME what about explicit/dds
         start_iteration = general.start_iteration
--- a/python/ngen_cal/src/ngen/cal/agent.py
+++ b/python/ngen_cal/src/ngen/cal/agent.py
@@ -85,6 +85,7 @@ class Agent(BaseAgent):
 
         if self._job is None:
             self._job = JobMeta(ngen_model.type, workdir, log=log)
+        ngen_model.workdir = self.job.workdir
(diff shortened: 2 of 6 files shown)


python/ngen_cal/src/ngen/cal/agent.py
```diff
@@ -57,9 +57,13 @@ def best_params(self) -> str:
 
 class Agent(BaseAgent):
 
-    def __init__(self, model_conf, workdir: 'Path', log: bool=False, restart: bool=False, parameters: 'Optional[Mapping[str, Any]]' = {}):
+    def __init__(self, model: Model, workdir: 'Path', log: bool=False, restart: bool=False, parameters: 'Optional[Mapping[str, Any]]' = {}):
         self._workdir = workdir
         self._job = None
+        assert not isinstance(model.model, NoModel), "invariant"
+        # NOTE: if support for new models is added, this will need to be modified
```
hellkite500, line 64: It isn't clear here exactly what will need to be modified under this Note

python/ngen_cal/src/ngen/cal/agent.py
```diff
@@ -57,9 +57,13 @@ def best_params(self) -> str:
...
+        model_inner = model.model.unwrap()
```
hellkite500, line 65: It may be worth a quick description of what the difference between `model_inner` and `model` is here

python/ngen_cal/src/ngen/cal/agent.py
```diff
@@ -70,17 +74,14 @@ def __init__(self, model_conf, workdir: 'Path', log: bool=False, restart: bool=F
...
-            workdirs = list(Path.glob(workdir, model_conf['type']+"_*_worker"))
+            workdirs = list(Path.glob(workdir, model_inner.type+"_*_worker"))
             if len(workdirs) > 1:
                 print("More than one existing workdir, cannot restart")
             elif len(workdirs) == 1:
-                self._job = JobMeta(model_conf['type'], workdir, workdirs[0], log=log)
+                self._job = JobMeta(model_inner.type, workdir, workdirs[0], log=log)
 
         if self._job is None:
-            self._job = JobMeta(model_conf['type'], workdir, log=log)
```
hellkite500, line 80: These steps are minutely important to the use of the Agent, and I don't think they can be dropped...The agent manipulates the configs to ensure root/workdirs are set appropriately under the automatically created agent working dirs.

hellkite500 (changes requested):
The change in behavior of the Agent and its interaction with the model config is quite likely to break.

python/ngen_cal/src/ngen/cal/agent.py
aaraney: Just that `model.model` could likely be more than `Ngen | NoModel`, so getting the `unwrap` call would likely need to change. Do you want me to add something about that?

python/ngen_cal/src/ngen/cal/agent.py
aaraney: Yeah that was a pretty awful name haha. Ill change that

python/ngen_cal/src/ngen/cal/agent.py
aaraney: I don't think they are dropped? I was just pulling them from the `pydantic` model instance now instead of the dictionary.

hellkite500: Sorry, this comment ended up being unclear, it was intended for lines 81-83 where workdir and binary are configured relative to the Agent's dynamically created workdir.  These are critical for correct operation of agents especially in PSO where multiple agents are running simultaneously.

As noted in disscussion with @aaraney the current `resolved_binary` isn't actually working as intended, and is its own bug which can be resolved in this context.

python/ngen_cal/src/ngen/cal/agent.py
```diff
@@ -72,17 +77,14 @@ def __init__(self, model_conf, workdir: Path, log: bool=False, restart: bool=Fal
...
             elif len(workdirs) == 1:
-                self._job = JobMeta(model_conf['type'], workdir, workdirs[0], log=log)
+                self._job = JobMeta(ngen_model.type, workdir, workdirs[0], log=log)
 
         if self._job is None:
-            self._job = JobMeta(model_conf['type'], workdir, log=log)
-        resolved_binary = Path(model_conf['binary']).resolve()
-        model_conf['workdir'] = self.job.workdir
-        self._model = Model(model=model_conf, binary=resolved_binary)
+            self._job = JobMeta(ngen_model.type, workdir, log=log)
```
hellkite500, line 87: I think we are still missing the updated model workdir.  An agent has a job which nests information and job/model files under a dynamically created workdir `job.workdir`.  Any model instance associated with that `job` must have its `workdir` aligned with the `job.workdir`, so we need to update the model reference like such
```suggestion
            self._job = JobMeta(ngen_model.type, workdir, log=log)
        ngen_model.workdir = self.job.workdir
```

python/ngen_cal/src/ngen/cal/__main__.py
```diff
@@ -42,13 +43,21 @@ def main(general: General, model_conf: Mapping[str, Any]):
...
+    model_inner = model.model.unwrap()
+
     plugins = cast(List[Union[Callable, ModuleType]], general.plugins)
     plugin_manager = setup_plugin_manager(plugins)
 
     print(_loaded_plugins(plugin_manager))
 
     # setup plugins
     plugin_manager.hook.ngen_cal_configure(config=general)
+    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
```
hellkite500, line 60: An agent effectively side effects the `Model` it associates with, particularly setting the `workdir` for that model run.  Because of this, we probably need to initialize the Agent before passing the Model to anything else, including plugins.
```suggestion
    # Initialize the starting agent
    # WARNING: Agent sets the model workdir, so initialize it before using the model object for anything that may require an accurate workdir
    agent = Agent(model, general.workdir, general.log, general.restart, general.strategy.parameters)
    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
```

hellkite500, line 60: An agent effectively side effects the `Model` it associates with, particularly setting the `workdir` for that model run.  Because of this, we probably need to initialize the Agent before passing the Model to anything else, including plugins.
```suggestion
    # Initialize the starting agent
    # WARNING: Agent sets the model workdir, so initialize it before using the model object for anything that may require an accurate workdir
    agent = Agent(model, general.workdir, general.log, general.restart, general.strategy.parameters)
    model_inner._plugin_manager.hook.ngen_cal_model_configure(config=model_inner)
```

python/ngen_cal/src/ngen/cal/__main__.py
```diff
@@ -57,8 +66,9 @@ def main(general: General, model_conf: Mapping[str, Any]):
     into a single variable vector and calibrating a set of heterogenous formultions...
     """
     start_iteration = 0
-    #Initialize the starting agent
-    agent = Agent(model_conf, general.workdir, general.log, general.restart, general.strategy.parameters)
+
+    # Initialize the starting agent
+    agent = Agent(model, general.workdir, general.log, general.restart, general.strategy.parameters)
```
hellkite500, line 71:
```suggestion
```

hellkite500 (approved):

PR #175, opened 2024-08-23
feat: add `topmodel` `init_config` models

aaraney:
Initial structures for parsing [`topmodel`](https://github.com/noaa-owp/topmodel) configuration files. Still a few things to sort so it is more compliant with `ngen.init_config`, but this is a good start.

### DIFF
--- a/python/ngen_conf/src/ngen/config/init_config/topmodel.py
+++ b/python/ngen_conf/src/ngen/config/init_config/topmodel.py
@@ -1,0 +1,464 @@
+from __future__ import annotations
+
+import io
+import pathlib
+import warnings
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    List,
+    Literal,
+    Protocol,
+    runtime_checkable,
+)
+
+import typing_extensions
+from pydantic import Field, root_validator, validator
+
+import ngen.init_config.serializer_deserializer as serde
+from ngen.config.path_pair import PathPair
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+
+@runtime_checkable
+class _Readliner(Protocol):
+    """
+    A type that implements a `readline` method.
+    For example a `io.TextIOWrapper` or `io.StringIO`:
+
+        with open("file.txt") as fp:
+            line = fp.readline()
+    """
+
+    def readline(self, size: int = -1, /) -> str: ...
+
+
+def _maybe_into_readliner(obj: Any) -> _Readliner | None:
+    if isinstance(obj, _Readliner):
+        return obj
+    elif isinstance(obj, str):
+        return io.StringIO(obj)
+    elif isinstance(obj, bytes):
+        # assume utf-8 encoding
+        obj = obj.decode()
+        return io.StringIO(obj)
+    else:
+        return None
+
+
+def _verify_topmodel_str(s: str, field_name: str) -> None:
+    assert (
+        len(s.encode()) < 256
+    ), f"`{field_name}` must be less that 256 bytes in length"
+
+
+class TopModelSubcat(serde.GenericSerializerDeserializer):
+    # 1 for ngen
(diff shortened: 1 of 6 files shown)


aaraney:
@hellkite500, there are a few assumptions that ive made here that we might want to change. In either case, this is ready for a first round of review.

python/ngen_conf/src/ngen/config/init_config/topmodel.py
```diff
@@ -0,0 +1,385 @@
...
+        assert distarea_lnaotb_str != "", "missing `dist_area_lnaotb` `lnaotb`"
+        distarea_lnaotb = distarea_lnaotb_str.split(" ")
+        assert (
+            len(distarea_lnaotb) == 2
+        ), f"incorrect number of fields in `dist_area_lnaotb` `lnaotb`, got {len(ntopo_area)}"
+        data["dist_area_lnaotb"] = float(distarea_lnaotb[0])
+        data["lnaotb"] = float(distarea_lnaotb[1])
+
+        num_channels_str = reader.readline()
+        assert distarea_lnaotb_str != "", "missing `num_channels`"
```
hellkite500, line 129: Wrong variable in check here, should be num channels

python/ngen_conf/src/ngen/config/init_config/topmodel.py
```diff
@@ -0,0 +1,385 @@
...
+    input: pathlib.Path = pathlib.Path("/dev/null")
+    subcat: PathPair[TopModelSubcat]
+    params: PathPair[TopModelParams]
+    output: pathlib.Path = pathlib.Path("/dev/null")
+    hyd: pathlib.Path = pathlib.Path("/dev/null")
+
+    @validator("stand_alone", pre=True)
+    @classmethod
+    def _coerce_stand_along(cls, value: str | int):
+        return int(value)
```
hellkite500, line 290: Can we warn if NGEN configs are marked with the standalone flag?

python/ngen_conf/src/ngen/config/init_config/topmodel.py
```diff
@@ -0,0 +1,385 @@
...
+        data["yes_print_output"] = int(cat_imap_out[2])
+
+        assert data["num_sub_catchments"] == 1, "only 1 subcatchment supported"
+
+        # drop trailing newline
+        title = reader.readline().rstrip("\r\n")
+        assert (
+            len(title.encode()) <= 256
+        ), "`title` must be less that 256 bytes in length"
+        data["subcat"] = title
```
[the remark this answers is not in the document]
aaraney: Yeah, I get what you mean. Lets meet in the middle and error if its `>255`. I would rather fail here if someone is using this for validation purposes and they are being really pedantic about the meaning of validation (never been me before 👀).

python/ngen_conf/src/ngen/config/init_config/topmodel.py
aaraney: Yeah, I think that's fair. I'd left it this way originally as I was implementing the bare minimum parser for NextGen. I'll go back and add the functionality to support more complex configuration files.

I think it's probably worth maintaining the complexity of a more complex parser and model for this module in particular. I envision TOPMODEL might be a candidate bmi module for a layered configuration in the future. What are your thoughts, @hellkite500?

hellkite500: I generally agree. And I think if we are building ngen specific configuration and validation, we should at least provide warnings of incompatible instances when we can.

hellkite500: I don't see it 👀

aaraney: Ooooops, you right. I put a warning somewhere else but not here. Just added it. My b.

PR #178, opened 2024-08-26
feat: add `Generic(De)Serializer` stub classes for generic formats

aaraney:
Add `GenericSerializer`, `GenericDeserializer`, and `GenericSerializerDeserializer` abstract marker classes. `ngen.init_config` cannot support all formats under the sun, but with other tools like `ngen.config_gen` being aware of the abstractions present in `ngen.init_config`, these classes provide a common interface for generic data formats to interoperate. Beyond interfaces, the `type`'s work as markers for other tools (e.g. `isinstance` check and change behavior).

## Additions

- Add `GenericSerializer`, `GenericDeserializer`, and `GenericSerializerDeserializer` abstract marker classes. Serializers and Deserializer must impliement the following methods respectively: `to_file(self, f: pathlib.Path, *_) -> None`, `to_str(self, *_) -> str` and `from_file(cls, f: pathlib.Path, *_)`, `from_str(cls, s: str, *_)`.


aaraney:
Lets merge this before #175. I am going to re-implement the work in #175 in terms of this.

python/ngen_init_config/src/ngen/init_config/deserializer.py
```diff
@@ -105,3 +107,21 @@ def from_json(cls, p: Path) -> Self:
...
+class GenericDeserializer(Base, ABC):
+    """
+    Stub for deserializing from a generic format.
+    Subclasses must implement the `from_file` and `from_str` abstract class methods.
+    See `ngen.init_config.core.Base` and `pydantic`'s documentation for configuration options.
+    """
+
+    @classmethod
+    @abstractmethod
+    def from_file(cls, p: Path, *_) -> Self:
```
aaraney, line 121: Note for reviewer, `*_` effectively means that subclasses that add arguments must make them keyword only / optional. The goal is improve uniformity of implementations.

hellkite500: I would add that note in the extended section of the doc string.

python/ngen_init_config/src/ngen/init_config/serializer_deserializer.py
```diff
@@ -70,3 +72,13 @@ class JsonSerializerDeserializer(ser.JsonSerializer, de.JsonDeserializer):
     """
 
     ...
+
+
+class GenericSerializerDeserializer(ser.GenericSerializer, de.GenericDeserializer, ABC):
+    """
```
hellkite500, line 78: Why does this class explicitly subclass `ABC`? It's effectively already an abstract class via inheritance, and it doesn't have any of it's own abstract methods. Other than possibly obscure MRO reasons is this needed?

aaraney: Just for clarity. It is redundant. My assumption is that most use cases will subclass this instead of `ser.GenericSerializer` or `de.GenericDeserializer` directly.  Unless you read the docs it might not be entirely clear that you need to implement methods. Happy to get rid of it if you think its confusing. Thoughts?

hellkite500: Python will tell you real quick if you don't implement them.  I would drop it.

aaraney: Just dropped it.

python/ngen_init_config/src/ngen/init_config/deserializer.py
hellkite500: I'm not too worried. I probably would literally have written it as "Note to self:" because I would forget in the future haha.

hellkite500 (approved):

PR #180, opened 2024-08-27
Shared param test

hellkite500:
Test to validate that the ngen module `_params_as_df` will ensure that any parameters with the same name end up mapping to the same permuted value.


python/ngen_cal/src/ngen/cal/ngen.py
```diff
@@ -46,13 +46,13 @@ def _params_as_df(params: Mapping[str, Parameters], name: str = None):
...
             dfs.append(df)
-        return pd.concat(dfs)
+        return pd.concat(dfs).set_index('param')
     else:
         p = params.get(name, [])
         df = pd.DataFrame([s.__dict__ for s in p])
         df['model'] = name
         df.rename(columns={'name':'param'}, inplace=True)
-        return df
+        return df.set_index('param')
```
hellkite500, line 55: Ugh.  It is actually possible for this dataframe to be empty if called from _map_params_to_realization and there is a module in the realization which doesn't have any params from the config to map to it, in this case, the `set_index` here will fail

hellkite500: Also, I think this breaks search, since `param` is no longer a column when subset to update_config

search.py", line 168, in dds_set

PR #182, opened 2024-08-28
Param mapping

hellkite500:
This should allow a calibration parameter to set an `alias` key, and any params across any models sharing the same `alias` name will be treated as the *same* parameter for the purpose of permutation generation.

Note, needs #180 to merge first.


python/ngen_cal/src/ngen/cal/parameter.py
```diff
@@ -11,5 +11,13 @@ class Parameter(BaseModel, allow_population_by_field_name = True):
     min: float
     max: float
     init: float
+    alias: Optional[str]
```
aaraney, line 14: Lets change this to the following since `alias` is non-`None`, we just rely on the validator to set `alias` to `name` if it was not provided. Just will avoid having to check if `alias` is `None` else where in the code to please static analysis tools.

```suggestion
    alias: str = None
    """Set to value of `name` if `alias` is not provided."""
```

python/ngen_cal/src/ngen/cal/parameter.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from pydantic import BaseModel, Field
-from typing import Sequence
+from pydantic import BaseModel, Field, root_validator
+from typing import Sequence, Mapping, Optional
 
 class Parameter(BaseModel, allow_population_by_field_name = True):
```
aaraney, line 6: In looking at the tests, what are your thoughts on validating that:
- `init`, `min`, and `max` are not `nan` (on the fence about this one)
- `init` is in the bounds of `min` and `max`
- `min` >= `max`? 

These seem like sane invariants that we _should_ uphold.

aaraney:
Aside, we should add a warning here if a `Parameter` mapping has been configured for a module that is not present.

https://github.com/NOAA-OWP/ngen-cal/blob/b7482dec44fda9e6cbf162729c72f4c1b8d9d623/python/ngen_cal/src/ngen/cal/ngen.py#L64

python/ngen_cal/tests/conftest.py
```diff
@@ -219,3 +222,32 @@ def explicit_catchments(nexus, fabric, workdir) -> Generator[ List[ CalibrationC
...
+    p1 = Parameter(name='a', min=0, max=1, init=0)
+    p2 = Parameter(name='d', min=2, max=3, init=0)
+    p3 = Parameter(name='c', min=0, max=1, init=0)
+    p4 = Parameter(name='a', min=0, max=1, init=0)
+    params = {'A':[p1], 'B':[p2, p4], 'C':[p3, p4]}
+
+    return params
+
+@pytest.fixture
+def multi_model_shared_params2() -> Mapping[str, list[Parameter]]:
```
aaraney, line 237:
```suggestion
def multi_model_shared_params_shared_name() -> Mapping[str, list[Parameter]]:
```

aaraney, line 237: At first glance, this overlaps with `multi_model_shared_params` s.t. I dont think we need it? Can we just get rid of this fixture?

python/ngen_cal/src/ngen/cal/parameter.py
```diff
@@ -11,5 +11,13 @@ class Parameter(BaseModel, allow_population_by_field_name = True):
     min: float
     max: float
     init: float
+    alias: Optional[str]
+
+    @root_validator
```
aaraney, line 16: Minor improvement that without could result in a misleading error.

```suggestion
    @root_validator(skip_on_failure=True)
```

PR #183, opened 2024-08-30
Fix `PathPair` & `PathPairCollection` validation;  add `path_pair` type hint fn; use `path_pair` in `Topmodel` `init_config` model

aaraney:
## Additions

- `path_pair` type hint fn. Use this instead of `PathPair` (see Feature Details section in #183)
- `Topmodel` `subcat` and `param` fields can now be initialized with `TopModelSubcat` and `TopModelParams` instances respectively.

## Changes

- Fix `PathPair` & `PathPairCollection` model validation
- Use `path_pair` type hint function in `Topmodel` `init_config` model
- `AbstractPathPairMixin.read` return type hint correct to `bool`. Already returned `bool`, type hint was missing.

## Feature Details

`path_pair` type hint function:

In nearly all cases, this should be used as the pydantic model field type if the field is to be a `PathPair[T]`. This will ensure that the `PathPair[T]` instance has the specified `serializer`, `deserializer`, `reader`, and `writer`.

NOTE: if a `PathPair` instance is passed to a model field defined using this function during initialization of the model, any `serializer`, `deserializer`, `reader`, or `writer` on the `PathPair` instance will not be replaced by the inputs provided to this function.

NOTE: if a pydantic model's field type is defined using this function, it's `ModelField.type_` will be a `PathPairOptions`.

Example:

```python
import typing
import pydantic
from ngen.config.path_pair import (
    PathPair,
    pydantic_serializer,
    pydantic_deserializer,
)

class Model(pydantic.BaseModel):
    field: int

class Foo(pydantic.BaseModel):
    # make static type checkers happy
    if typing.TYPE_CHECKING:
        path: PathPair[Model]
    else:
        path: path_pair(
            Model,
            serializer=pydantic_serializer,
            deserializer=pydantic_deserializer(Model),
        )
```

### DIFF
--- a/python/ngen_conf/src/ngen/config/path_pair/path_pair.py
+++ b/python/ngen_conf/src/ngen/config/path_pair/path_pair.py
@@ -136,6 +136,101 @@ class PathPair(AbstractPathPairMixin[T], Path, Generic[T]):
         return cls(path_validator(value))
 
 
+# NOTE: this is an established pattern pydantic uses for 'constrained' types.
+# like . See pydantic constrained types (e.g. `pydantic.types.ContstrainedInt`
+# and `conint`) for other examples of this pattern.
+def path_pair(
+    t: type[T],
+    *,
+    serializer: Serializer[T] | None = None,
+    deserializer: Deserializer[T] | None = None,
+    reader: Reader = path_reader,
+    writer: Writer = path_writer,
+) -> type[PathPairOptions]:
+    """
+    In nearly all cases, this should be used as the pydantic model field type
+    if the field is to be a `PathPair[T]`. This will ensure that the
+    `PathPair[T]` instance has the specified `serializer`, `deserializer`,
+    `reader`, and `writer`.
+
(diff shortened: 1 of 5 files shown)


python/ngen_conf/src/ngen/config/path_pair/path_pair.py
```diff
@@ -82,11 +122,113 @@ def __modify_schema__(cls, field_schema: dict[str, Any]):
...
+
+    @classmethod
+    def __modify_schema__(cls, field_schema: dict[str, Any]) -> None:
+        field_schema.update(PathPair[cls.ty].schema())  # type: ignore
 
-        return PathPair(value)
+    @classmethod
+    def __get_validators__(cls) -> CallableGenerator:
+        yield from PathPair[cls.ty].__get_validators__()
+        # apply all validators, now `value` is guaranteed to be a `PathPair`
```
hellkite500, line 213: which `value`?

hellkite500 (approved):

PR #194, opened 2024-09-09
Param hotfix

hellkite500:
Previously, only single model parameter spaces were fully supported by ngen-cal.

This *feature* makes it possible to calibrate a multi-bmi model with ngen by allowing multiple parameter sets from different models to be combined into a single search space with a unique (ish) index.  

> [!NOTE]
> Parameters with the _exact_ same name between two models will be exposed as two parameters in the parameter space, but will always share the same value.

### DIFF
--- a/python/ngen_cal/src/ngen/cal/ngen.py
+++ b/python/ngen_cal/src/ngen/cal/ngen.py
@@ -46,13 +46,22 @@ def _params_as_df(params: Mapping[str, Parameters], name: str = None):
             df['model'] = k
             df.rename(columns={'name':'param'}, inplace=True)
             dfs.append(df)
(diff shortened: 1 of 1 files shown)


aaraney:
I don't think this works at least for `dds` in the uniform case b.c. this does not guarantee unique index values. When `dds_update` selects a random subset of parameters, if two of the parameters share the same name (e.g. 2 bmi modules have calib param with same name) the following code throw.

https://github.com/NOAA-OWP/ngen-cal/blob/0f5843d713e76fb0a23eef6e25e4d75d789b2f7d/python/ngen_cal/src/ngen/cal/search.py#L77-L94

aaraney (approved):

PR #187, opened 2024-09-04
Update `cfe`, `topmodel`, and `lgar` calibratable parmeters; `ngen.cal` python < 3.9 pin `fiona<1.10.0`

aaraney:
## Additions

- `LgarParams`
- docstrings for `CfeParams`, `TopmodelParams`, and `LgarParams` parameters
- New `CfeParams` and `TopmodelParams` calibratable parameters

## Removals

- `CfeParams` `bb` and `multiplier` fields dropped. `bb` is now `b`; support for `multiplier` was dropped in https://github.com/NOAA-OWP/cfe/pull/64

## Changes

- `ngen.cal` on `python` `3.9` pins `fiona<1.10.0` (see notes)

## Notes:

`fiona` `v1.10.0` deprecated  `fiona.path` field (seem to have been removed?) [see](https://github.com/Toblerity/Fiona/releases/tag/1.10.0).

`geopandas` `geojson` reader uses `fiona`.

`geopandas` `v0.14` dropped support for `python<3.9`. So, `ngen.cal` installations are effectively pinned to `geopandas<=0.14`. Transitively, `ngen.cal` needs to pin `fiona<1.10.0` for compatibility.
https://geopandas.org/en/stable/docs/changelog.html#version-0-14-sep-15-2023

PR #195, opened 2024-09-10
fix(py3.12): explicitly call `__init__` to setup `PurePath` instance

### LINKED ISSUE #193: `PathPair` is not printable in `3.12`; Explicitly call `__init__` after instance has been established
https://github.com/NOAA-OWP/ngen-cal/blob/0f5843d713e76fb0a23eef6e25e4d75d789b2f7d/python/ngen_conf/src/ngen/config/path_pair/path_pair.py#L86

```patch
diff --git a/python/ngen_conf/src/ngen/config/path_pair/path_pair.py b/python/ngen_conf/src/ngen/config/path_pair/path_pair.py
index f829676..57275ab 100644
--- a/python/ngen_conf/src/ngen/config/path_pair/path_pair.py
+++ b/python/ngen_conf/src/ngen/config/path_pair/path_pair.py
@@ -84,6 +84,8 @@ class PathPair(AbstractPathPairMixin[T], Path, Generic[T]):
             inner = _MaybeInner[cls._parameters[0]](inner=inner).inner  # type: ignore
         cls = WindowsPathPair[T] if os.name == "nt" else PosixPathPair[T]
         self: WindowsPathPair[T] | PosixPathPair[T] = Path.__new__(cls, *args, **kwargs)
+        # explicitly call __init__ to setup instance
+        self.__init__(*args, **kwargs)
         self._inner = inner  # type: ignore
         self._serializer = serializer  # type: ignore
         self._deserializer = deserializer  # type: ignore
```

A change in `pathlib.PurePath` introduced in `python` `3.12` seems to cause an issue when trying to call `__str__` (or `__repr__`) on a `PathPair` instance. The face value solution is to explicitly "setup" the `PosixPathPath` or `WindowsPathPair` instance by calling it's `__init__` method. I need to investigate why this is the case document any repercussions of this solution.

aaraney:
Fixes #193.

See #193 for detail.

### DIFF
--- a/.github/workflows/ngen-cal.yaml
+++ b/.github/workflows/ngen-cal.yaml
@@ -15,11 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        #3.10 ends up building a lot of python deps as whls (e.g. pandas) and it takes a LONG time (11m42s)
-        #so for now, don't enable testing on 3.10.  3.11 similarly builds some deps, but it
-        #particaularly fails to build the tables dep since the ubuntu image doesn't have hdf5
-        #When 3.11 tables wheel is available, then add it?  Also, put 3.10 in quotes or it turns into 3.1
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


aaraney:
Once #187 is merged, this needs to be rebased off of `master` and re-pushed. This should fix the test failures.

hellkite500 (approved):

Issue #210, opened 2024-12-04
Update to v2.2 hydrofabric data model


JordanLaserGit:
Some of the variable names in the `divide-attributes` layer (was `model-attributes` <v2.2) have changed in the new [v.2.2 hydrofabric data model](https://lynker-spatial.s3-us-west-2.amazonaws.com/hydrofabric/v2.2/hfv2.2-data_model.html). This is causing ngen-cal to error when calling generate_configs() for CFE if the hf file is v2.2. 

Examples of variable name changes

"bexp_soil_layers_stag=1" is now "mode.bexp_soil_layers_stag=1"

and 

"dksat_soil_layers_stag=1" is now "geom_mean.dksat_soil_layers_stag=1"

PR #211, opened 2024-12-05
update variable names to v2.2 data model

JordanLaserGit:
Updates ngen-cal to use the new v2.2 hydrofabric data model. Merging this PR will make ngen-cal incompatible with earlier data model versions. We can add checks to support backwards compatibility if desired. This addresses issue https://github.com/NOAA-OWP/ngen-cal/issues/210.

## Additions
None
-

## Removals
None
-

## Changes
example:
"bexp_soil_layers_stag=1" -> "mode.bexp_soil_layers_stag=1"
-

## Testing

1. 3 tests are failing, but do not appear to be related to these changes

### DIFF
--- a/python/ngen_config_gen/src/ngen/config_gen/models/cfe.py
+++ b/python/ngen_config_gen/src/ngen/config_gen/models/cfe.py
@@ -65,8 +65,8 @@ class Cfe:
 
         # saturated hydraulic conductivity
         # NOTE: it seems all values each layer of `dksat_soil_layers_stag` are the same.
-        x_data = data.get("dksat_soil_layers_stag=1",None)
-        if x_data is None: x_data = data.get("geom_mean.dksat_soil_layers_stag=1",None)
+        x_data = data.get("geom_mean.dksat_soil_layers_stag=1",None)
+        if x_data is None: x_data = data.get("dksat_soil_layers_stag=1",None)
         if x_data is None: raise Exception(f"Neither geom_mean.dksat_soil_layers_stag=1 nor dksat_soil_layers_stag=1 were found in attributes")        
         self.data["soil_params_satdk"] = FloatUnitPair(
             value=x_data,
@@ -101,7 +101,7 @@ class Cfe:
 
         # wilting point soil moisture content
         # NOTE: it seems all values each layer of `smcwlt_soil_layers_stag` are the same.
-        x_data = data.get("mean.smcwlt_soil_layers_stag=1=1",None)
+        x_data = data.get("mean.smcwlt_soil_layers_stag=1",None)
         if x_data is None: x_data = data.get("smcwlt_soil_layers_stag=1",None)
         if x_data is None: raise Exception(f"Neither mean.smcwlt_soil_layers_stag=1 nor smcwlt_soil_layers_stag=1 were found in attributes")         
         self.data["soil_params_wltsmc"] = FloatUnitPair(
@@ -122,9 +122,9 @@ class Cfe:
         self.data["cgw"] = FloatUnitPair(value=x_data, unit=M_PER_H)
 
         # exponent parameter (1.0 for linear reservoir)
-        x_data = data.get("mean.Expon",None)
+        x_data = data.get("mode.Expon",None)
         if x_data is None: x_data = data.get("gw_Expon",None)
-        if x_data is None: raise Exception(f"Neither mean.Expon nor gw_Expon were found in attributes")         
+        if x_data is None: raise Exception(f"Neither mode.Expon nor gw_Expon were found in attributes")         
         self.data["expon"] = FloatUnitPair(value=x_data, unit=EMPTY)
 
     def _v2_defaults(self) -> None:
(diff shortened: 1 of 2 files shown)


hellkite500:
Until 2.2 is more widely adopted, I think we should probably check/support both, marking 2.1 support for deprecation.

JordanLaserGit:
Checks have been added. User will get an exception if the neither variable option is found in the attributes.

aaraney:
Thanks for writing this up @JordanLaserGit! I envision that we will face similar issues when future versions of the HF are released. Likewise, this seems pretty crosscutting where I would expect many BMI config generation scripts need to be updated to accommodate the HF 2.2 `model-attributes` schema changes. For that reason, I prefer going with crosswalk or deferred mapping approach that `ngen.config_gen` could provide as a utility instead of just updating existing config generation scripts. Below is an example of kind of what I had in mind. Will you take a look at it and let me know what you think? If you can think of a simpler way to accomplish the same thing, i'm all ears!

```python
# crosswalk between HF v2 model attribute names and HF v2.2 divide attributes
# names. 
import typing


_v2_to_v2_2_mapping = {
    "aspect_c_mean": "circ_mean.aspect",
    "bexp_soil_layers_stag=1": "mode.bexp_soil_layers_stag=1",
    "bexp_soil_layers_stag=2": "mode.bexp_soil_layers_stag=2",
    "bexp_soil_layers_stag=3": "mode.bexp_soil_layers_stag=3",
    "bexp_soil_layers_stag=4": "mode.bexp_soil_layers_stag=4",
    "cwpvt": "mean.cwpvt",
    "divide_id": "divide_id",
    "dksat_soil_layers_stag=1": "geom_mean.dksat_soil_layers_stag=1",
    "dksat_soil_layers_stag=2": "geom_mean.dksat_soil_layers_stag=2",
    "dksat_soil_layers_stag=3": "geom_mean.dksat_soil_layers_stag=3",
    "dksat_soil_layers_stag=4": "geom_mean.dksat_soil_layers_stag=4",
    "elevation_mean": "mean.elevation",
    "gw_Coeff": "mean.Coeff",
    "gw_Expon": "mode.Expon",
    "gw_Zmax": "mean.Zmax",
    "impervious_mean": "mean.impervious",
    "ISLTYP": "mode.ISLTYP",
    "IVGTYP": "mode.IVGTYP",
    "mfsno": "mean.mfsno",
    "mp": "mean.mp",
    "psisat_soil_layers_stag=1": "geom_mean.psisat_soil_layers_stag=1",
    "psisat_soil_layers_stag=2": "geom_mean.psisat_soil_layers_stag=2",
    "psisat_soil_layers_stag=3": "geom_mean.psisat_soil_layers_stag=3",
    "psisat_soil_layers_stag=4": "geom_mean.psisat_soil_layers_stag=4",
    # NOTE: aaraney: missing from 2.2?
    # "quartz_soil_layers_stag=1":
    # "quartz_soil_layers_stag=2":
    # "quartz_soil_layers_stag=3":
    # "quartz_soil_layers_stag=4":
    "refkdt": "mean.refkdt",
    "slope_mean": "mean.slope",
    "slope": "mean.slope_1km",
    "smcmax_soil_layers_stag=1": "mean.smcmax_soil_layers_stag=1",
    "smcmax_soil_layers_stag=2": "mean.smcmax_soil_layers_stag=2",
    "smcmax_soil_layers_stag=3": "mean.smcmax_soil_layers_stag=3",
    "smcmax_soil_layers_stag=4": "mean.smcmax_soil_layers_stag=4",
    "smcwlt_soil_layers_stag=1": "mean.smcwlt_soil_layers_stag=1",
    "smcwlt_soil_layers_stag=2": "mean.smcwlt_soil_layers_stag=2",
    "smcwlt_soil_layers_stag=3": "mean.smcwlt_soil_layers_stag=3",
    "smcwlt_soil_layers_stag=4": "mean.smcwlt_soil_layers_stag=4",
    "twi_dist_4": "dist_4.twi",
    "vcmx25": "mean.vcmx25",
    "X": "centroid_x",
    "Y": "centroid_y",
}


def divide_attributes_v2_to_v2_2_crosswalk(d: dict[str, typing.Any]):
    class Translate:
        def __contains__(self, name: str) -> bool:
            return name in _v2_to_v2_2_mapping or name in d

        def __getitem__(self, name: str) -> typing.Any:
            if name in _v2_to_v2_2_mapping:
                v2_2_name = _v2_to_v2_2_mapping[name]
                return d[v2_2_name]
            # fallback
            return d[name]

    return Translate()
```

To use this, you would pass the `data` parameter from `hydrofabric_linked_data_hook` to   `divide_attributes_v2_to_v2_2_crosswalk` and it would return something you could treat like the `data` dictionary, but handle the remapping if necessary. This should let us and other users wrap the `data` dictionary at the top of their `hydrofabric_linked_data_hook` implementation and not need to modify any other existing code.

JordanLaserGit:
@aaraney Sure this looks like a fine solution. Will need to write new mappings if the variables change again, but shouldn't be too tough and would need to do that with my solution anyway. I went with the quick hack just to get ngen-cal to play nice with the new HF version, but the differed mapping solution will keep the code cleaner and serves as a reference for how the variables changed from one version to the next.

PR #218, opened 2025-03-31
fix: correct a set of v2.1 model attribute names for config mapping

hellkite500:
Fix a few parameter name mappings


hellkite500:
I realized that the intended mapping was 2.2 -> 2.0

The proposed changes here support 2.2 -> 2.1, so probably shouldn't be merged as is.

Leaving this open for now to show differences though.

PR #219, opened 2025-04-07
fix: temp solution to allow externally define noah-owp-modular params

hellkite500:
quick workaround so that parameters passed into a noah-owp-modular model persist into the object.

### DIFF
--- a/python/ngen_conf/src/ngen/config/noahowp.py
+++ b/python/ngen_conf/src/ngen/config/noahowp.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel, Field
 from .bmi_formulation import BMIFortran
 
 
-class NoahOWPParams(BaseModel):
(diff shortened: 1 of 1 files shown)


aaraney:
I don't have a huge problem with this, just curious why the downstream caller can't just subclass `NoahOWPParams` and enable `extra="allow"`?

hellkite500:
Because the realization parsing uses this parameter definition, but as is, NOM parameters will always be empty since none are defined. This simply allows the realization/formulation models to capture params when deserializing a Realization, for example.

The only way to do that from the caller would be monkey patching I think...

aaraney:
> The only way to do that from the caller would be monkey patching I think...

Yeah that is kind of annoying. I think you could do something like `NoahOWPParams.Config.extra = "allow"` but if we are using this in a program like `ngen.cal`, then yeah that will not work.

aaraney (approved):

